PR #8, opened 2017-11-22
Make connection termination configurable 

pubudu91:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/TargetHandler.java
```diff
@@ -184,4 +198,10 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
             }
         }
     }
+
+    private void closeChannel(ChannelHandlerContext ctx) throws Exception {
+        if (ctx.channel().isActive()) {
```
shafreenAnfar: Can we explain the reason for this if condition

pubudu91: Ack.

PR #12, opened 2017-11-29
Remove content-Type for empty payloads and add content length zero

chamil321:

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
@@ -154,17 +154,21 @@ public class Util {
     }
 
     private static void  setContentLength(HTTPCarbonMessage cMsg) {
-        if (cMsg.isAlreadyRead() || (cMsg.getHeader(Constants.HTTP_CONTENT_LENGTH) == null && !cMsg.isEmpty())) {
+        if (cMsg.isAlreadyRead() || (cMsg.getHeader(Constants.HTTP_CONTENT_LENGTH) == null && (!cMsg.isEmpty()
+                || cMsg.getProperty(Constants.HTTP_METHOD).toString().equals(Constants.HTTP_POST_METHOD)))) {
             Util.prepareBuiltMessageForTransfer(cMsg);
             int contentLength = cMsg.getFullMessageLength();
             if (contentLength > 0) {
                 cMsg.setHeader(Constants.HTTP_CONTENT_LENGTH, String.valueOf(contentLength));
+            } else {
+                cMsg.setHeader(Constants.HTTP_CONTENT_LENGTH, String.valueOf(0));
             }
         }
     }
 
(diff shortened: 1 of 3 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
```diff
@@ -154,17 +154,21 @@ public static void setupTransferEncodingForRequest(HTTPCarbonMessage cMsg, boole
...
+            } else {
+                cMsg.setHeader(Constants.HTTP_CONTENT_LENGTH, String.valueOf(0));
             }
         }
     }
 
     private static void  setTransferEncodingHeader(HTTPCarbonMessage cMsg) {
-        if (cMsg.isAlreadyRead() || (cMsg.getHeader(Constants.HTTP_TRANSFER_ENCODING) == null && !cMsg.isEmpty())) {
+        if (cMsg.isAlreadyRead() || (cMsg.getHeader(Constants.HTTP_TRANSFER_ENCODING) == null && (!cMsg.isEmpty()
+                || cMsg.getProperty(Constants.HTTP_METHOD).toString().equals(Constants.HTTP_POST_METHOD)))) {
```
shafreenAnfar: Are we only checking for POST ? what about PUT, PATCH and DELETE ?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
```diff
@@ -159,6 +159,8 @@ private static void  setContentLength(HTTPCarbonMessage cMsg) {
             int contentLength = cMsg.getFullMessageLength();
             if (contentLength > 0) {
                 cMsg.setHeader(Constants.HTTP_CONTENT_LENGTH, String.valueOf(contentLength));
+            } else if (isMethodAllowed(cMsg.getProperty(Constants.HTTP_METHOD).toString())) {
```
shafreenAnfar: shall we changed the name of the function. Maybe we can say isEntityBodyAllowed ?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
```diff
@@ -167,21 +169,27 @@ private static void  setTransferEncodingHeader(HTTPCarbonMessage cMsg) {
...
                     return;
                 }
             }
             cMsg.setHeader(Constants.HTTP_TRANSFER_ENCODING, Constants.CHUNKED);
         }
     }
 
+    private static boolean isMethodAllowed(String method) {
+        return method.equals(Constants.HTTP_POST_METHOD) || method.equals(Constants.HTTP_PUT_METHOD)
+                || method.equals(Constants.HTTP_PATCH_METHOD) || method.equals(Constants.HTTP_DELETE_METHOD);
```
shafreenAnfar: We need to take out DELETE verb here. See below, 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE

PR #13, opened 2017-11-29
Set the connection:close header when keep-alive is disabled

pubudu91:
Ubuntu Gnome 16.04
tools: cURL, Chrome 60.0.3112.78


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpClientConnectorImpl.java
```diff
@@ -93,6 +94,9 @@ public void operationComplete(ChannelFuture channelFuture) throws Exception {
                             setChannelAttributes(channelFuture.channel(), httpCarbonRequest, httpResponseFuture,
                                                  targetChannel);
                         }
+                        if (!keepAlive) {
+                            httpCarbonRequest.setHeader(Constants.CONNECTION, Constants.CONNECTION_CLOSE);
+                        }
                         targetChannel.writeContent(httpCarbonRequest);
```
shafreenAnfar: Can you move line [1] to line 99 ? Just before we write only we need to set that.

[1] https://github.com/wso2/transport-http/pull/13/files#diff-f76950f6ceaa84347012883dbd6dcd2cR92

PR #17, opened 2017-12-11
Remove all the data-source related code from transport

shafreenAnfar:
Hence, we have removed it from the transport.

## Automation tests
 - Covered by existing ones.

## Security checks
 - N/A

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/BlockingEntityCollector.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/BlockingEntityCollector.java
@@ -43,15 +43,15 @@ public class BlockingEntityCollector implements EntityCollector {
     private static final Logger LOG = LoggerFactory.getLogger(BlockingEntityCollector.class);
 
     private int soTimeOut;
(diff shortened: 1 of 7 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/BlockingEntityCollector.java
```diff
@@ -84,23 +97,31 @@ public HttpContent getHttpContent() {
...
+                    isEndOfMessageProcessed = true;
+                }
+                contentList.add(httpContent);
 
-    public ByteBuf getMessageBody() {
-        HttpContent httpContent = getHttpContent();
-        if (httpContent != null) {
-            return httpContent.content();
+            } catch (InterruptedException e) {
+                LOG.error("Error while getting full message length", e);
```
isudana: should we just log and ignore this ?

shafreenAnfar: Yeah. Because in case this gets interrupted, we retry on the next loop.

PR #26, opened 2017-12-20
Add testcase for connection re-use in connection-pool

shafreenAnfar:
In addition to that I have refactored the constant names that are used in Test-Cases.  

## Automation tests
 - Available

## Security checks
 - N/A


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolTestCase.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @AfterClass
+    public void cleanUp() throws ServerConnectorException {
+        TestUtil.cleanUp(serverConnectors, httpServer);
+    }
+
+    private class ClientWorker implements Callable<String> {
+
+        String response;
```
irunika: Shouldn't we make this private?

PR #29, opened 2017-12-22
Update Netty version and refactor deprecated Netty API usage

pubudu91:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/ServerConnectorFuture.java
```diff
@@ -136,6 +136,7 @@
      *
      * @param serverConnectorId The ID of the server connected related to this port unbinding event
      * @param isHttps Specifies whether the server connector is using HTTPS.
+     * @throws ServerConnectorException Thrown if there is an error in unbinding the port.
```
irunika: IMO we can remove "thrown" here. WDYT?

PR #31, opened 2017-12-26
Remove singleton ConnectionManager and add testcases

shafreenAnfar:
> Hence, I thought of making ConnectionManager per client-connector. While doing so I came across few other singletons which I have removed with this PR.

> In addition to that, When the client connector is used without any association with the server connector, connection pool is not reusing the connections. This has been fixed with this PR as well.

## Automation tests
 - Added

## Security checks
 - N/A

### DIFF
--- a/components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolMainTestCase.java
+++ b/components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolMainTestCase.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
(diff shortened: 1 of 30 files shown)


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolMainTestCase.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private HTTPConnectorListener sendRequestAsync(CountDownLatch latch) {
+        HTTPCarbonMessage httpsPostReq = TestUtil.
+                createHttpsPostReq(TestUtil.HTTP_SERVER_PORT, "hello", "/");
+        HTTPConnectorListener requestListener = new HTTPConnectorListener(latch);
+        HttpResponseFuture responseFuture = httpClientConnector.send(httpsPostReq);
+        responseFuture.setHttpConnectorListener(requestListener);
+        return requestListener;
+    }
+
```
anupama-pathirage: Can we move the both above methods (waitAndGetStringEntity and sendRequestAsync) to a common place like TestUtil class as this code blocks are duplicated in both ConnectionPoolMainTestCase class and ConnectionPoolEvictionTestCase class.

PR #32, opened 2018-01-04
Add test case for max active connections configuration

pubudu91:


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolMaxConnTestCase.java
```diff
@@ -0,0 +1,93 @@
+package org.wso2.transport.http.netty.connectionpool;
```
shafreenAnfar: Seems the license header is missing..

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolMaxConnTestCase.java
```diff
@@ -0,0 +1,93 @@
...
+public class ConnectionPoolMaxConnTestCase {
+
+    private static final Logger log = LoggerFactory.getLogger(ConnectionPoolMaxConnTestCase.class);
+
+    private HttpServer httpServer;
+    private HttpClientConnector httpClientConnector;
+
+    @BeforeClass
+    public void setup() {
+        TransportsConfiguration transportsConfiguration = TestUtil.getConfiguration(
```
shafreenAnfar: Here the vertical affinity is wrong. TransportConfiguration is needed by the client connector not by the httpServer. So let's move it further down the code.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolMaxConnTestCase.java
shafreenAnfar: And also shall we try writing the testcase by extending ChunkClientTemplate.java

pubudu91: Added the license header.. Cannot use ChunkClientTemplate.java since we cannot send requests asynchronously in it.

PR #36, opened 2018-01-11
Refactor the code

shafreenAnfar:
## Automation tests
 - Covered by existing testcases

## Security checks
 - N/A

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
@@ -44,7 +44,7 @@ import java.util.Locale;
  */
 public class HttpOutboundRespListener implements HttpConnectorListener {
 
-    private static final Logger log = LoggerFactory.getLogger(HttpConnectorListener.class);
+    private static final Logger log = LoggerFactory.getLogger(HttpOutboundRespListener.class);
 
     private ChannelHandlerContext sourceContext;
     private RequestDataHolder requestDataHolder;
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
@@ -24,6 +24,8 @@ import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.http.HttpContent;
 import io.netty.handler.codec.http.HttpResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.wso2.transport.http.netty.common.Constants;
 import org.wso2.transport.http.netty.common.Util;
 import org.wso2.transport.http.netty.config.ChunkConfig;
@@ -35,11 +37,14 @@ import org.wso2.transport.http.netty.message.HTTPCarbonMessage;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
 /**
  * Get executed when the response is available.
  */
 public class HttpOutboundRespListener implements HttpConnectorListener {
+
+    private static final Logger log = LoggerFactory.getLogger(HttpConnectorListener.class);
 
     private ChannelHandlerContext sourceContext;
     private RequestDataHolder requestDataHolder;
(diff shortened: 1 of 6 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
```diff
@@ -35,12 +37,15 @@
...
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
 /**
  * Get executed when the response is available.
  */
 public class HttpOutboundRespListener implements HttpConnectorListener {
 
+    private static final Logger log = LoggerFactory.getLogger(HttpConnectorListener.class);
```
pubudu91: Any particular reason to use the super class for the logger instead of the actual class?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
```diff
@@ -71,58 +76,79 @@ public void onMessage(HTTPCarbonMessage outboundResponseMsg) {
...
-                                && Util.isVersionCompatibleForChunking(requestDataHolder.getHttpVersion())) {
-                            Util.setupChunkedRequest(outboundResponseMsg);
-                        } else {
-                            contentLength += httpContent.content().readableBytes();
-                            Util.setupContentLengthRequest(outboundResponseMsg, contentLength);
+                        try {
+                            writeOutboundResponse(outboundResponseMsg, keepAlive, httpContent);
+                        } catch (Exception exception) {
+                            String errorMsg = "Failed to send the outbound response : "
+                                    + exception.getMessage().toLowerCase(Locale.ENGLISH);
```
pubudu91: Do the potential exceptions which can end up here have meaningful messages? If not, it'd be better to handle each of those cases separately to ensure the error message logged is meaningful.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
```diff
@@ -71,58 +76,79 @@ public void onMessage(HTTPCarbonMessage outboundResponseMsg) {
...
-                    if (handlerExecutor != null) {
-                        handlerExecutor.executeAtSourceResponseSending(outboundResponseMsg);
-                    }
-                    outboundResponseMsg.removeHttpContentAsyncFuture();
-                    contentList.clear();
-                    contentLength = 0;
+    private void writeOutboundResponse(HTTPCarbonMessage outboundResponseMsg, boolean keepAlive,
+            HttpContent httpContent) {
+        if (Util.isLastHttpContent(httpContent)) {
+            if (!isHeaderWritten) {
```
pubudu91: Can we change the name of `isHeaderWritten`? When reading this, got the wrong impression that it's about a header related to chunking, when in fact it refers to whether all the headers of the response were written to the channel or not.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
shafreenAnfar: It was a mistake and fixed it. Thanks for pointing out :)

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
pubudu91: Another slightly related thing: I noticed that `isHeaderWritten` is set in `writeOutboundResponseHeaders()` method. In it, `isHeaderWritten` is set to true before writing the response to the channel. Shouldn't it be set to true once we know if it was successfully written?

For example, something like below:
```java
private void writeOutboundResponseHeaders(HTTPCarbonMessage httpOutboundRequest, boolean keepAlive) {
        HttpResponse response = Util.createHttpResponse(httpOutboundRequest, keepAlive);
        sourceContext.write(response).addListener( listener -> this.isHeaderWritten = true );
}
```

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
shafreenAnfar: At this point we don't really expect any run-time errors to thrown as those are reflected in the future result. That is something we have to address separately. This exception block is just the last line defense to capture and notify any unexpected exceptions.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
shafreenAnfar: Chunks does not really have headers that are interesting to users. Chunking headers are only used by decoders. If you look look at the method name it sets context for isHeaderWritten variable. Hence, I think it is okay to leave it as it is.

PR #39, opened 2018-01-22
Make transport more resilient to failures

shafreenAnfar:
- Client closes the connection while sending request headers. 
- Client closes the connection after sending request headers. 
- Client closes the connection while sending request body. 
- Client closes the connection while receiving request headers/body. 
- Client suddenly stops sending the request. 

Same scenarios were mirrored for server side as well.  

## Automation tests
 - Unit tests 
   > Couldn't write automation test for this as it is difficult mimic these scenarios. Specially, we need a proper TCP client and server. Hence, differed it for the moment and created an issue to track the requirement. https://github.com/ballerinalang/ballerina/issues/4475


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultOperationStatus.java
```diff
@@ -0,0 +1,26 @@
+package org.wso2.transport.http.netty.contractimpl;
```
anupama-pathirage: Add license header for this

shafreenAnfar: Done.

PR #40, opened 2018-01-24
Multipart request handling for server connector

arukshani:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/HttpBodyPart.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Represents one body part of a multipart message.
+ */
+public class HttpBodyPart implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    private final byte[] content;
+    private final String contentType;
+    private final String partName;
+    private final String fileName;
```
anupama-pathirage: fileName seems unused. Do we need this?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/MultipartRequestDecoder.java
```diff
@@ -0,0 +1,164 @@
...
+                LOG.warn("Received unknown attribute type.");
+                break;
+        }
+        multiparts.add(bodyPart);
+    }
+
+    /**
+     * Reset request decoder.
+     */
+    public void resetPostRequestDecoder() {
```
anupama-pathirage: Shall we make this method private

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/HttpBodyPart.java
arukshani: This is going to be needed in ballerina side.

PR #48, opened 2018-02-14
Add support for handling compressing request and decompressing response

shafreenAnfar:
## Automation tests
 - New test-cases are added. 

## Security checks
 - No

### DIFF
--- a/components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/compression/ServerRespCompressionTestCase.java
+++ b/components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/compression/ServerRespCompressionTestCase.java
@@ -47,7 +47,7 @@ import java.util.HashMap;
 import static org.testng.AssertJUnit.assertEquals;
 
 /**
- * This class tests for 414 and 413 responses.
+ * This class tests compression outbound responses.
  */
 public class ServerRespCompressionTestCase {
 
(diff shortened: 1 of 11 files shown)


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/compression/ServerRespCompressionTestCase.java
```diff
@@ -0,0 +1,122 @@
...
+            httpRequest.headers().set(Constants.ACCEPT_ENCODING, Constants.ENCODING_DEFLATE);
+            httpResponse = httpClient.sendRequest(httpRequest);
+            assertEquals(Constants.ENCODING_DEFLATE, httpResponse.headers().get(HttpHeaderNames.CONTENT_ENCODING));
+
+        } catch (Exception e) {
+            TestUtil.handleException("IOException occurred while running serverConnectorRespCompressionTest", e);
+        }
+    }
+
+    @AfterClass
```
Manuri: If I remember correctly in order to run this method even if a test(s) fails/skips, we have to explicitly set "alwaysRun=true"

shafreenAnfar: @Manuri I read the doc but didn't fully understand what they mean. Is there a way to try out this functionality ?

http://testng.org/doc/documentation-main.html

shafreenAnfar: As discussed I tried throwing a RuntimeException as well as failing the test-case but still couldn't reproduce the issue.

Manuri: The docs say "If set to true, this configuration method will be run even if one or more methods invoked previously failed or was skipped". However I couldn't find the default value in the doc. If the cleanup properly happens then may be the default is alwaysRun=true. If thats the case I think we can keep this as it is.

PR #58, opened 2018-02-21
Fix some bugs

shafreenAnfar:
1) NPE when returning TargetChannel to pool.
2) Creating EventLoopGroups for each connection creation.
3) Not returning the channel when timesout.

## Automation tests
 - Unit tests 
   > Yes

## Security checks
 > N/A


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/SourceHandler.java
```diff
@@ -58,18 +58,17 @@
  */
 public class SourceHandler extends ChannelInboundHandlerAdapter {
     private static Logger log = LoggerFactory.getLogger(SourceHandler.class);
-    private static final String HTTP_RESOURCE = "httpResource";
 
-    protected ChannelHandlerContext ctx;
-    protected HTTPCarbonMessage sourceReqCmsg;
-    protected HandlerExecutor handlerExecutor;
+    HTTPCarbonMessage sourceReqCmsg;
+    HandlerExecutor handlerExecutor;
```
anupama-pathirage: Shall we make above 2 protected?

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/util/server/initializers/SendChannelActiveCountServerInitializer.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
anupama-pathirage: Incorrect year?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/SourceHandler.java
shafreenAnfar: Intellij suggested package-private. And for me also that seems correct.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/HttpResponseFuture.java
```diff
@@ -57,7 +57,7 @@
     /**
      * Let make the async operation sync.
      * @return Status future
-     * @throws InterruptedException
+     * @throws InterruptedException throws when interrupted
```
irunika: Since throws in already in the doc can we change the message like "@throws InterruptedException if the connection is interrupted" ?

PR #59, opened 2018-02-21
Use lower case headers consistently

riyafa:

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
@@ -91,7 +91,6 @@ public final class Constants {
     public static final String TRUSTSTOREFILE = "trustStoreFile";
     public static final String TRUSTSTOREPASS = "trustStorePass";
 
-    public static final String PORT = "PORT";
 
     public static final int DEFAULT_HTTP_PORT = 80;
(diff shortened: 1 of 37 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/ResponseContentWriter.java
```diff
@@ -59,7 +59,7 @@ public void writeLastContent(CarbonMessage carbonMessage) {
 //            HTTPTransportContextHolder.getInstance().getHandlerExecutor().
 //                    executeAtSourceResponseSending(carbonMessage);
```
lankavitharana: please remove these commented lines.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
```diff
@@ -109,19 +108,9 @@
     public static final String ENCODING_DEFLATE = "deflate";
     public static final String HTTP_TRANSFER_ENCODING_IDENTITY = "identity";
 
-    // TODO: Move string constants for HTTP headers and header values to their own class
-    public static final String HTTP_CONNECTION = "Connection";
+    // TODO: Move string constants for HTTP header values to their own class
```
anupama-pathirage: Can we remove this todo now?

PR #60, opened 2018-02-23
Fix keep-alive config not working

riyafa:
## Automation tests
 - Unit tests 
   > Added in transport-http

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/SourceHandler.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/SourceHandler.java
@@ -39,6 +39,7 @@ import org.slf4j.LoggerFactory;
 import org.wso2.transport.http.netty.common.Constants;
 import org.wso2.transport.http.netty.common.Util;
 import org.wso2.transport.http.netty.config.ChunkConfig;
+import org.wso2.transport.http.netty.config.KeepAliveConfig;
 import org.wso2.transport.http.netty.contract.HttpResponseFuture;
 import org.wso2.transport.http.netty.contract.ServerConnectorException;
 import org.wso2.transport.http.netty.contract.ServerConnectorFuture;
@@ -65,6 +66,7 @@ public class SourceHandler extends ChannelInboundHandlerAdapter {
     private Map<String, GenericObjectPool> targetChannelPool;
     private ServerConnectorFuture serverConnectorFuture;
     private ChunkConfig chunkConfig;
+    private KeepAliveConfig keepAliveConfig;
     private HttpResponseFuture httpOutboundRespFuture;
     private String interfaceId;
     private String serverName;
@@ -73,10 +75,11 @@ public class SourceHandler extends ChannelInboundHandlerAdapter {
     protected ChannelHandlerContext ctx;
 
     public SourceHandler(ServerConnectorFuture serverConnectorFuture, String interfaceId, ChunkConfig chunkConfig,
-            String serverName, ChannelGroup allChannels) {
+                         KeepAliveConfig keepAliveConfig, String serverName, ChannelGroup allChannels) {
         this.serverConnectorFuture = serverConnectorFuture;
         this.interfaceId = interfaceId;
         this.chunkConfig = chunkConfig;
+        this.keepAliveConfig = keepAliveConfig;
         this.targetChannelPool = new ConcurrentHashMap<>();
         this.idleTimeout = false;
         this.serverName = serverName;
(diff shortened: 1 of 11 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/config/ChunkKeepAliveConfig.java
```diff
@@ -22,7 +22,7 @@
 /**
  * Contains value for configuring content encoding
  */
-public enum ChunkConfig {
+public enum ChunkKeepAliveConfig {
```
pubudu91: Why was this renamed? The `ChunkConfig` is for configuring chunking. Keep-alive configurations shouldn't be mixed in with it IMO

riyafa: I am using the same enum for both cases. Otherwise need to write the same enum with a different name.

pubudu91:
@riyafa If using the same enum for both, then let's make its name a generic one. `ChunkKeepAliveConfig` just sounds weird. Also, can you explain how the keep-alive configuration map to these enum values? In what scenario does `AUTO` apply to keep-alive config?

riyafa:
`AUTO` is for the default value, `ALWAYS` is to enable keep-alive and `NEVER` is to disable keep-alive

pubudu91:
@riyafa But what is the default value? keep-alive is a boolean value. You either keep the connection or you close it.

riyafa:
Default value is different for HTTP 1.0 and 1.1. 
Default for HTTP 1.0 is `close` and for HTTP 1.1 is `keep-alive`

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/http1point0test/HttpOnePointZeroServerConnectorTestCase.java
```diff
@@ -83,6 +80,7 @@ public void setUp() {
     @Test
     public void http1point0DefaultRequest() {
         try {
+            setUp();
```
shafreenAnfar: Can't we use @BeforeMethod ?

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/http1point0test/HttpOnePointZeroServerConnectorTestCase.java
```diff
@@ -92,32 +90,30 @@ public void http1point0DefaultRequest() {
...
             assertNotNull(httpResponse.headers().get(Constants.HTTP_CONTENT_LENGTH));
+            serverConnector.stop();
         } catch (Exception e) {
-            TestUtil.handleException("IOException occurred while running largeHeaderTest", e);
+            TestUtil.handleException("IOException occurred while running http1point0KeepAliveRequest test", e);
         }
     }
-
-    @AfterClass
-    public void cleanUp() throws ServerConnectorException {
```
shafreenAnfar: Here also cant we use @Aftermethod ?

shafreenAnfar: Also we need to shutdown httpWsConnectorFactory as well..

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/http1point0test/HttpOnePointZeroServerConnectorTestCase.java
riyafa: We cannot in this test because we need to have different `listenerConfiguration`s for each test case.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/SourceHandler.java
```diff
@@ -180,8 +184,9 @@ void notifyRequestListener(HTTPCarbonMessage httpRequestMsg, ChannelHandlerConte
             try {
                 ServerConnectorFuture outboundRespFuture = httpRequestMsg.getHttpResponseFuture();
                 outboundRespFuture
-                        .setHttpConnectorListener(new HttpOutboundRespListener(ctx, httpRequestMsg,
-                                chunkConfig, serverName));
+                        .setHttpConnectorListener(
+                                new HttpOutboundRespListener(ctx, httpRequestMsg, chunkConfig, keepAliveConfig,
+                                                             serverName));
```
anupama-pathirage: Shall we reformat this line ?

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/http1point0test/HttpOnePointZeroServerConnectorTestCase.java
riyafa: `httpWsConnectorFactory` doesn't have a method to shut it down

riyafa: Fixed

PR #64, opened 2018-02-26
[HTTP] Add Test Cases for Waiting Timeout for Idle Connection in Connection Pool

irunika:
This PR adds test cases for connection timeout when a connection is waiting for an idle connection in a connection pool.
This also fixes the issue of not having a connection related error message when connection waiting timeout occurs.


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpClientConnectorImpl.java
```diff
@@ -140,6 +142,10 @@ private void notifyErrorState(ChannelFuture channelFuture) {
                 }
             });
         } catch (Exception failedCause) {
+            if (failedCause instanceof NoSuchElementException
+                    && "Timeout waiting for idle object".equals(failedCause.getMessage())) {
+                failedCause = new NoSuchElementException(Constants.TIMEOUT_WAITING_FOR_IDLE_CONNECTION);
```
pubudu91: Is this for avoiding exceptions with Java level info/implementation info in their error messages? If so, I think we don't need to worry about those at the transport level.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
```diff
@@ -270,6 +270,8 @@
     public static final String REMOTE_SERVER_CLOSE_RESPONSE_CONNECTION_AFTER_REQUEST_READ
             = "Remote host closed the connection without sending inbound response";
 
+    public static final String TIMEOUT_WAITING_FOR_IDLE_CONNECTION = "Timeout waiting for idle connection";
```
pubudu91: Can we rephrase this? Not sure whether this gives the correct message to the user.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolWaitingTimeoutTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+
+            // Check the responses.
+            Throwable throwable = null;
+            HashSet<String> channelIds = new HashSet<>();
+            for (HTTPConnectorListener responseListener : responseListeners) {
+                if (responseListener.getHttpErrorMessage() != null) {
+                    if (throwable != null) {
+                        Assert.assertTrue(false, "Cannot have more than one error");
+                    }
+                    throwable = responseListener.getHttpErrorMessage();
```
pubudu91: Why is this a single variable?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpClientConnectorImpl.java
irunika: It is generic exception coming from the API. But we only throw this error from the HTTP client connector. So IMO it is better if we can return a more specific error in this level.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolWaitingTimeoutTestCase.java
irunika: This case has been written for maximum connection in the pool to be two. Other request returns back with the error since it cannot obtain a connection from the pool within maximum wait time. If needed this can have a array of values.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
irunika: Done

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
```diff
@@ -270,6 +270,9 @@
     public static final String REMOTE_SERVER_CLOSE_RESPONSE_CONNECTION_AFTER_REQUEST_READ
             = "Remote host closed the connection without sending inbound response";
 
+    public static final String MAXIMUM_WAIT_TIME_EXCEED =
+            "Maximum wait time exceeds. Cannot obtain a connection to send the request.";
```
shafreenAnfar: shall we say "Cannot obtain a connection withing maximum wait time"

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolWaitingTimeoutTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class ConnectionPoolWaitingTimeoutTestCase {
+
+    private HttpServer httpServer;
+    private HttpClientConnector httpClientConnector;
+    private HttpWsConnectorFactory connectorFactory;
+
+    private static final int MAX_ACTIVE_CONNECTIONS = 2;
+    private static final int MAX_WAIT_TIME_FOR_CONNECTION_POOL = 1000;
+
```
shafreenAnfar: Shall we remove white spaces ?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
irunika: Yes. That makes more sense. Will change it.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolWaitingTimeoutTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+            Assert.assertTrue(throwable instanceof NoSuchElementException);
+            Assert.assertEquals(throwable.getMessage(), Constants.MAXIMUM_WAIT_TIME_EXCEED);
+        } catch (Exception e) {
+            TestUtil.handleException("IOException occurred while running testMaxActiveConnectionsPerPool", e);
+        }
+    }
+
+    @AfterClass
+    public void cleanUp() throws ServerConnectorException {
+        TestUtil.cleanUp(new ArrayList<>(), httpServer);
```
shafreenAnfar: We need to shutdown connectorFactory as well.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolWaitingTimeoutTestCase.java
shafreenAnfar: I guess it is because only one client could get the error.

irunika: Yes. From all three clients only one client should get the error.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
```diff
@@ -270,6 +270,9 @@
     public static final String REMOTE_SERVER_CLOSE_RESPONSE_CONNECTION_AFTER_REQUEST_READ
             = "Remote host closed the connection without sending inbound response";
 
+    public static final String MAXIMUM_WAIT_TIME_EXCEED =
+            "Cannot obtain a connection within maximum wait time";
```
shafreenAnfar: Sorry instead of "Cannot" shall we "Could not"

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolWaitingTimeoutTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+        senderConfiguration.getPoolConfiguration().setMaxWaitTime(MAX_WAIT_TIME_FOR_CONNECTION_POOL);
+        httpClientConnector = connectorFactory.createHttpClientConnector(new HashMap<>(), senderConfiguration);
+    }
+
+    @Test
+    public void testWaitingForConnectionTimeout() {
+        try {
+            int noOfRequests = 3;
+
+            // Create countdown latches
```
shafreenAnfar: This comment is not needed right ? Remember comments are only necessary evil :)

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolWaitingTimeoutTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+            CountDownLatch[] countDownLatches = new CountDownLatch[noOfRequests];
+            for (int i = 0; i < noOfRequests; i++) {
+                countDownLatches[i] = new CountDownLatch(1);
+            }
+
+            // Send multiple requests asynchronously to force the creation of multiple connections
```
shafreenAnfar: This is also not needed :)

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/connectionpool/ConnectionPoolWaitingTimeoutTestCase.java
irunika: Got it.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
irunika: Sure

PR #68, opened 2018-02-28
Implement OCSP stapling.

Bhashinee:
## Purpose
Support OCSP stapling feature. 

## Approach
OCSP stapling is the nowadays used protocol for verifying whether a certificate is revoked or not during the SSL handshake. But still some legacy systems do not support this. Because of that by default this will be disabled. 

## Security checks

Related Issue: https://github.com/ballerina-lang/ballerina/issues/3856


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/OCSPResponseBuilder.java
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * A class for generating OCSP response.
+ */
+public class OCSPResponseBuilder {
+
+    public static OCSPResp getOcspResponse(SSLConfig sslConfig, int cacheAllcatedSize, int cacheDelay)
+            throws IOException, KeyStoreException, UnrecoverableEntryException, NoSuchAlgorithmException,
+            CertificateVerificationException {
+
```
Manuri: Can we break this method down into smaller units?

Bhashinee: Moved as requested.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/certificatevalidation/ocsp/OCSPCache.java
```diff
@@ -171,9 +171,24 @@ public synchronized SingleResp getCacheValue(BigInteger serialNumber) {
...
+        OCSPCacheValue cacheValue = hashMap.get(serialNumber);
+        if (cacheValue != null) {
+            if (!cacheValue.isValid()) {
+                cacheValue.updateCacheWithNewValue();
+                OCSPCacheValue ocspCacheValue = hashMap.get(serialNumber);
+                return (ocspCacheValue != null ? ocspCacheValue.getOCSPValue() : null);
+            }
+
+            return cacheValue.getOCSPValue();
+        } else {
```
chamil321: Shall we get rid of this else by null checking the cacheValue at the begining?
if (cacheValue == null) {
return null;
}

PR #71, opened 2018-03-04
Add configuration for access log

riyafa:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/ServerConnectorBootstrap.java
```diff
@@ -148,6 +148,9 @@ public void addHttpTraceLogHandler(Boolean isHttpTraceLogEnabled) {
         httpServerChannelInitializer.setHttpTraceLogEnabled(isHttpTraceLogEnabled);
     }
 
+    public void addHttpAccessLogHandler(Boolean isHttpTraceLogEnabled) {
```
chamil321: Shall we rename the argument name? isHttpAccessLogEnabled*

PR #77, opened 2018-03-05
Introduce Timeout exception

VijithaEkanayake:
When an endpoint timeout occurred, the listener will be notified with a client connector exception. this will introduce EndpointTimeoutError which will be helpful to implement programs more resilient to failures.


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/passthrough/PassthroughMessageProcessorListener.java
```diff
@@ -82,7 +83,8 @@ public void onMessage(HTTPCarbonMessage httpResponse) {
 
                     @Override
                     public void onError(Throwable throwable) {
-                        if (throwable instanceof ClientConnectorException) {
+                        if (throwable instanceof ClientConnectorException
+                                || throwable instanceof EndpointTimeOutException) {
```
shafreenAnfar: Do we need || here ?

VijithaEkanayake: Yes, we don't here.

PR #81, opened 2018-03-10
Add ALPN support for HTTP2 over TLS

Bhashinee:
## Purpose
HTTP/2 over TLS must use ALPN(Application Layer Protocol Negotiation) to avoid additional network round trips by including protocol negotiation in the TLS handshake phase. Purpose is to incorporate this support.

## Goals
Add support for client and server to use ALPN extension in http2 connection.

## Approach
As part of the initial Client hello, the client provides a list of protocols which it supports. The server then selects a protocol from that list and sends the selected protocol as part of the server hello. This is happening in TLS handshake phase. For previously established connections, the protocol selection can be negotiated over an abbreviated hand shake.

## Security checks

Related Issue : https://github.com/ballerina-lang/ballerina/issues/4831


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/ConnectionAvailabilityFuture.java
```diff
@@ -0,0 +1,87 @@
...
+        channelFuture.addListener(new ChannelFutureListener() {
+
+            @Override
+            public void operationComplete(ChannelFuture channelFuture) throws Exception {
+                if (isValidChannel(channelFuture)) {
+                    socketAvailable = true;
+                    if (listener != null && !isSSLEnabled) {
+                        notifyListener(Constants.HTTP_SCHEME);
+                    }
+                } else {
```
isudana: If the listener is not set at this moment  there will be a NPE. 
We need to store that there is an error, and notify when listener is set later.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/ConnectionAvailabilityFuture.java
```diff
@@ -0,0 +1,93 @@
...
+            notifyListener(protocol);
+        } else if (!isSSLEnabled && socketAvailable) {
+            notifyListener(Constants.HTTP_SCHEME);
+        } else if (isFailure) {
+            listener.onFailure(channelFuture);
+        }
+    }
+
+    public void setFailure() {
+        listener.onFailure(channelFuture);
```
isudana: if listener is not set, there will be a NPE here as well.

PR #87, opened 2018-03-15
Fix the null pointer issue in target handler

Bhashinee:
## Purpose
Fix the null pointer exception which comes when an error occurs.

## Approach
When an error occurs, in a TLS handshake it was giving a null pointer error since the target handler was not set at that moment.
As a solution, removed the rest of the handlers(including target handler) from the pipeline after SSLHandler if the handshake fails.

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
@@ -202,7 +202,7 @@ public final class Constants {
 
     public static final String HTTP_SOURCE_HANDLER = "SourceHandler";
     public static final String HTTP_ENCODER = "encoder";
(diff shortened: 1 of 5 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
```diff
@@ -202,7 +202,7 @@
 
     public static final String HTTP_SOURCE_HANDLER = "SourceHandler";
     public static final String HTTP_ENCODER = "encoder";
-    public static final String HTTP_DECODER = "decoder";
+    public static final String HTTP_CLIENT_CODEC = "decoder";
```
isudana: Client codec contains both encoder and decoder, so shall we rename the constant value ?

Bhashinee: Done

PR #100, opened 2018-04-09
Add proper SSL engine for redirect handler

arukshani:
Add proper SSL engine for redirect handler


arukshani:
Added a test for https redirect

PR #110, opened 2018-04-11
Use Enum for Keep-Alive Config in Sender

VijithaEkanayake:
## Purpose
This will introduce enum for keep-alive configuration in sender configuration.


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/TargetHandler.java
```diff
@@ -283,4 +292,21 @@ public HttpResponseFuture getHttpResponseFuture() {
...
+                if (Float.valueOf((String) getIncomingMsg().getProperty(Constants.HTTP_VERSION)) > Constants.HTTP_1_0) {
+                    return true;
+                } else {
+                    return false;
+                }
+            case ALWAYS:
+                return true;
+            case NEVER:
+                return false;
+            default:
```
chamil321: Can we remove this default case if we are certain about possible enums?

VijithaEkanayake: The execution will never reach here. To make the code compilable, default case has to be placed here.

PR #118, opened 2018-04-17
[WebSocket] Test Cases to Test Connection closure of WebSocket Client

irunika:
This PR resolves https://github.com/wso2/transport-http/issues/113

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketClient.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketClient.java
@@ -125,6 +125,7 @@ public class WebSocketClient {
                             pipeline.addLast(sslCtx.newHandler(ch.alloc(), host, port));
                         }
                         pipeline.addLast(new HttpClientCodec());
+                        // Assuming that WebSocket Handshake messages will not be large than 8KB
                         pipeline.addLast(new HttpObjectAggregator(8192));
                         pipeline.addLast(WebSocketClientCompressionHandler.INSTANCE);
                         if (idleTimeout > 0) {


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/internal/websocket/WebSocketSessionImpl.java
```diff
@@ -54,6 +54,7 @@ public WebSocketSessionImpl(ChannelHandlerContext ctx, boolean isSecure, String
                                 String sessionId) throws URISyntaxException {
         this.ctx = ctx;
         this.isSecure = isSecure;
+        this.isOpen = true;
```
shafreenAnfar: Can we change the name of this class from WebSocketSessionImpl to DefaultWebSocketSession.

shafreenAnfar: Also instead of saying isOpen shall we say open.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketClient.java
```diff
@@ -116,25 +116,24 @@ public HandshakeFuture handshake() {
...
+                new ChannelInitializer<SocketChannel>() {
+                    @Override
+                    protected void initChannel(SocketChannel ch) {
+                        ChannelPipeline pipeline = ch.pipeline();
+                        if (sslCtx != null) {
+                            pipeline.addLast(sslCtx.newHandler(ch.alloc(), host, port));
                         }
-                    });
+                        pipeline.addLast(new HttpClientCodec());
+                        pipeline.addLast(new HttpObjectAggregator(8192));
```
shafreenAnfar: Please add comment explaining why we went with 8192

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketTargetHandler.java
```diff
@@ -107,12 +107,12 @@ public void channelActive(ChannelHandlerContext ctx) throws URISyntaxException {
 
     @Override
     public void channelInactive(ChannelHandlerContext ctx) throws ServerConnectorException {
-        if (channelSession != null) {
+        if (channelSession != null && channelSession.isOpen()) {
             channelSession.setIsOpen(false);
+            int statusCode = 1001;
```
shafreenAnfar: Can't we have constants for these status codes. Netty itself might have those.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketTargetHandler.java
```diff
@@ -107,12 +107,12 @@ public void channelActive(ChannelHandlerContext ctx) throws URISyntaxException {
...
+            String reasonText = "Server is going away";
```
shafreenAnfar: I guess same applies for this as well.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/util/server/websocket/WebSocketRemoteServerFrameHandler.java
```diff
@@ -51,13 +53,20 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
             String text = ((TextWebSocketFrame) frame).text();
-            if (PING.equals(text)) {
-                ctx.writeAndFlush(new PingWebSocketFrame(Unpooled.wrappedBuffer(new byte[]{1, 2, 3, 4})));
-                return;
+            switch (text) {
+                case PING:
+                    ctx.writeAndFlush(new PingWebSocketFrame(Unpooled.wrappedBuffer(new byte[]{1, 2, 3, 4})));
+                    break;
+                case CLOSE:
+                    ctx.writeAndFlush(new CloseWebSocketFrame(1000, "Close on request"));
```
shafreenAnfar: Shall we use proper constants for 1000

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketTargetHandler.java
```diff
@@ -107,12 +107,11 @@ public void channelActive(ChannelHandlerContext ctx) throws URISyntaxException {
 
     @Override
     public void channelInactive(ChannelHandlerContext ctx) throws ServerConnectorException {
-        if (channelSession != null) {
+        if (channelSession != null && channelSession.isOpen()) {
             channelSession.setIsOpen(false);
+            String reasonText = "Server is going away";
+            notifyCloseMessage(Constants.WEBSOCKET_STATUS_CODE_GOING_AWAY, reasonText, ctx);
```
riyafa: What do you think of having the reason text as an empty string since the status code gives the meaning?

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/websocket/WebSocketTestClientConnectorListener.java
```diff
@@ -178,4 +182,27 @@ public boolean isIdleTimeout() throws Throwable {
        throw errorsQueue.remove();
     }
 
+    /**
+     * Return whether the connection is closed or not.
+     *
+     * @return true if the connection is closed.
+     * @throws Throwable if any error occurred.
+     */
+    public boolean isClose() throws Throwable {
```
riyafa: Shall we change this function name to isClosed?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketTargetHandler.java
irunika: 1001 is only saying "Going away". It does not say "what" is going away according to https://tools.ietf.org/html/rfc6455#section-7.4
So I thought saying what is going away is better. 
WDYT?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/internal/websocket/WebSocketSessionImpl.java
irunika: Done

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketTargetHandler.java
riyafa: I am suggesting that if user needs to set a reason text he should do it in code. There is not really server client concept in WebSocket after connection establishment. Is there? When we were using WebSocket in the browser and we close the browser then we receive a close but the reason text would be empty string.

PR #127, opened 2018-04-20
[WebSocket] Throttle Read WebSocket Frames

irunika:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/websocket/WsClientConnectorConfig.java
```diff
@@ -32,17 +32,19 @@
     private final String remoteAddress;
     private List<String> subProtocols;
     private int idleTimeoutInSeconds = -1;
+    private boolean autoRead = true;
     private final Map<String, String> headers = new HashMap<>();
```
shafreenAnfar: Shall we initialize all the variables inside the constructor instead of doing at the class load time ?

Issue #135, opened 2018-04-22
[WebSocket] Cannot handle continuation frames


irunika:
**Description:**
When an incoming message is not a final fragment of a large message WebSocket frame set a flag "fin" to indicate that it is not a final message. Then it sends continuation frames until the final message. This is not handled in current transport.

PR #136, opened 2018-04-22
[WebSocket] Add support for Continuation WebSocket Frames

irunika:
This resolves https://github.com/wso2/transport-http/issues/135

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
@@ -70,7 +70,7 @@ public class WebSocketSourceHandler extends ChannelInboundHandlerAdapter {
     private final String interfaceId;
     private String subProtocol = null;
     private HandlerExecutor handlerExecutor;
-    private WebSocketFrameType frameType = WebSocketFrameType.TEXT;
+    private WebSocketFrameType continuationFrameType;
 
     /**
      * @param connectorFuture {@link ServerConnectorFuture} to notify messages to application.
@@ -159,16 +159,19 @@ public class WebSocketSourceHandler extends ChannelInboundHandlerAdapter {
             logger.error("Expecting WebSocketFrame. Unknown type.");
             throw new UnknownWebSocketFrameTypeException("Expecting WebSocketFrame. Unknown type.");
         }
+
+        // If the continuation of frames are not following the protocol, netty handles it internally.
+        // Because of that those situations are not handles here.
         if (msg instanceof TextWebSocketFrame) {
             TextWebSocketFrame textFrame = (TextWebSocketFrame) msg;
             if (!textFrame.isFinalFragment()) {
-                frameType = WebSocketFrameType.TEXT;
+                continuationFrameType = WebSocketFrameType.TEXT;
             }
             notifyTextMessage(textFrame, textFrame.text(), textFrame.isFinalFragment());
         } else if (msg instanceof BinaryWebSocketFrame) {
             BinaryWebSocketFrame binaryFrame = (BinaryWebSocketFrame) msg;
             if (!binaryFrame.isFinalFragment()) {
-                frameType = WebSocketFrameType.BINARY;
+                continuationFrameType = WebSocketFrameType.BINARY;
             }
             notifyBinaryMessage(binaryFrame, binaryFrame.content(), binaryFrame.isFinalFragment());
         } else if (msg instanceof CloseWebSocketFrame) {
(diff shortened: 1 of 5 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
```diff
@@ -155,27 +159,50 @@ public void channelRead(ChannelHandlerContext ctx, Object msg)
             logger.error("Expecting WebSocketFrame. Unknown type.");
             throw new UnknownWebSocketFrameTypeException("Expecting WebSocketFrame. Unknown type.");
         }
+
+        // If the continuation of frames are not following the protocol, netty handles it internally.
+        // Because of that those situations are not handles here.
```
riyafa: Shall we change?
```
// If the continuation of frames are not following the protocol, netty handles them internally.
 // Hence those situations are not handled here.
```

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketTargetHandler.java
```diff
@@ -147,11 +151,22 @@ public void channelRead(ChannelHandlerContext ctx, Object msg)
                     "Unexpected FullHttpResponse (getStatus=" + response.status() +
                             ", content=" + response.content().toString(CharsetUtil.UTF_8) + ')');
         }
+
+        // If the continuation of frames are not following the protocol, netty handles it internally.
+        // Because of that those situations are not handles here.
```
riyafa: same here update comment

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
```diff
@@ -155,27 +159,50 @@ public void channelRead(ChannelHandlerContext ctx, Object msg)
...
         if (msg instanceof TextWebSocketFrame) {
-            notifyTextMessage((TextWebSocketFrame) msg);
+            TextWebSocketFrame textFrame = (TextWebSocketFrame) msg;
+            if (!textFrame.isFinalFragment()) {
+                continuationFrameType = WebSocketFrameType.TEXT;
+            }
+            notifyTextMessage(textFrame, textFrame.text(), textFrame.isFinalFragment());
```
riyafa: Only trying to understand. Will the `TextWebSocketFrame` come before a `ContinuationWebSocketFrame` ? Then do we first set the `continuationFrameType` and then notify it and then notify the continuationframes afterwards?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
irunika: To what?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketTargetHandler.java
irunika: To what?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
irunika: It does not come every time. 
If WebSocketFrame -> isFinalFragment = false 
then next frame should be a continuation frame. Errors of this sequence are already handled by netty in transport level. 

I actually cannot understand your suggestion. Next frame will not hit the channel read until one frame processing is finished.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
riyafa: to this: 
```
// If the continuation of frames are not following the protocol, netty handles them internally.
 // Hence those situations are not handled here.
```

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
riyafa: Got it thanks

irunika: :)

PR #146, opened 2018-04-25
Fix WebSocket Connection Closure Issue

irunika:
This PR addresses the transport related part of https://github.com/ballerina-platform/ballerina-lang/issues/7893

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
@@ -210,8 +210,6 @@ public class WebSocketSourceHandler extends ChannelInboundHandlerAdapter {
     private void notifyCloseMessage(CloseWebSocketFrame closeWebSocketFrame) throws ServerConnectorException {
         String reasonText = closeWebSocketFrame.reasonText();
         int statusCode = closeWebSocketFrame.statusCode();
-        ctx.channel().close();
-        webSocketConnection.getDefaultWebSocketSession().setIsOpen(false);
         WebSocketMessageImpl webSocketCloseMessage = new WebSocketCloseMessageImpl(statusCode, reasonText);
         closeWebSocketFrame.release();
         setupCommonProperties(webSocketCloseMessage);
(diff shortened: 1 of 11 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
```diff
@@ -138,17 +138,13 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
...
             handlerExecutor = null;
         }
 
-        if (webSocketConnection.getDefaultWebSocketSession().isOpen()) {
-            webSocketConnection.getDefaultWebSocketSession().setIsOpen(false);
-            int statusCode = 1001; // Client is going away.
-            String reasonText = "Client is going away";
-            notifyCloseMessage(statusCode, reasonText);
+        if (webSocketConnection != null) {
+            notifyCloseMessage(-1, null);
```
riyafa: Why -1?

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/util/client/websocket/WebSocketClientHandler.java
```diff
@@ -123,7 +119,15 @@ public void channelRead0(ChannelHandlerContext ctx, Object msg) throws Exception
             bufferReceived = pongFrame.content().nioBuffer();
         } else if (frame instanceof CloseWebSocketFrame) {
             logger.debug("WebSocket Client received closing");
-            isOpen = false;
+            if (ch.isOpen()) {
```
shafreenAnfar: Instead of ch can't say channel ?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/websocket/WebSocketTargetHandler.java
```diff
@@ -277,6 +269,8 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
         if (!handshakeFuture.isDone()) {
             handshakeFuture.setFailure(cause);
         }
+
+        //TODO: Should we close the connection without close frame?
```
riyafa: When channel is inactive need to send 1006 and close the connection.

irunika: I put this comment to remember about onError event which we have not designed yet. I think we need to broadly think about it and do it.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/WebSocketSourceHandler.java
irunika: In the current implementation, there is no way of telling the application which the connection is closed without close frame.
1006 is a good choise. But since this is not the end of the close discussion let's put it in the next release after a discussion. :)

PR #158, opened 2018-05-11
Fix incorrect cancelling of the handshake

riyafa:
- Restricts the status error status codes
 - Returns a channel future

 Further the makes sure that the `WebSocketInitMessage` returns the proper `HttpCarbonRequest` instead of the `HttpRequest`.

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/websocket/WebSocketInitMessage.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/websocket/WebSocketInitMessage.java
@@ -19,6 +19,7 @@
 
 package org.wso2.transport.http.netty.contract.websocket;
 
+import io.netty.channel.ChannelFuture;
 import io.netty.handler.codec.http.HttpHeaders;
 
 /**
@@ -90,8 +91,10 @@ public interface WebSocketInitMessage extends WebSocketMessage {
      *
      * @param closeCode close code for cancelling the handshake.
      * @param closeReason reason for canceling the handshake.
+     *
+     * @return the ChannelPromise created after submitting response
      */
-    void cancelHandShake(int closeCode, String closeReason);
+    ChannelFuture cancelHandshake(int closeCode, String closeReason);
 
     /**
      * Check whether the handshake is cancelled in someplace or not.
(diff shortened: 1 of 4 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/websocket/WebSocketInitMessage.java
```diff
@@ -91,7 +92,7 @@ HandshakeFuture handshake(String[] subProtocols, boolean allowExtensions, int id
      * @param closeCode close code for cancelling the handshake.
      * @param closeReason reason for canceling the handshake.
      */
-    void cancelHandShake(int closeCode, String closeReason);
+    ChannelPromise cancelHandshake(int closeCode, String closeReason);
```
irunika: Why are we returning ChannelPromise instead of ChannelFuture?

riyafa: Because we need to close the channel after responding successfully and it is not possible to get the channel from `ChannelFuture` without casting it.

irunika: Why should we give the channel to the user? Shouldn't we close the channel after sending the response to the client? In all the places we have used ChannelFuture because of the same reason. By giving ChannelPromise, users have the capability to deal with the channel and even our own pipeline.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/message/DefaultWebSocketInitMessage.java
```diff
@@ -115,24 +117,29 @@ public HandshakeFuture handshake(String[] subProtocols, boolean allowExtensions,
...
                 }
-            });
-        } finally {
-            isCancelled = true;
+                return responseFuture;
+            } finally {
+                isCancelled = true;
+            }
+        } else {
+            throw new IllegalStateException(
```
irunika: Isn't it good to throw two errors depending on `isCancelled` & `handshakeStarted`? So the user can refer to the exact problem for the failure?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/message/DefaultWebSocketInitMessage.java
```diff
@@ -115,24 +116,32 @@ public HandshakeFuture handshake(String[] subProtocols, boolean allowExtensions,
...
     @Override
-    public void cancelHandShake(int statusCode, String closeReason) {
-        try {
-            ChannelFuture responseFuture;
-            if (closeReason != null) {
-                ByteBuf content = Unpooled.wrappedBuffer(closeReason.getBytes(StandardCharsets.UTF_8));
-                responseFuture = ctx.writeAndFlush(new DefaultFullHttpResponse(HttpVersion.HTTP_1_1,
-                                                                      HttpResponseStatus.valueOf(statusCode), content));
+    public ChannelFuture cancelHandshake(int statusCode, String closeReason) {
+        if (!isCancelled && !handshakeStarted) {
```
shafreenAnfar: Instead of isCancelled shall we cancelled ?

PR #170, opened 2018-05-22
Do OCSP and CRL validation if the server fails to respond with ocsp staple

Bhashinee:
## Purpose
Fix the issue : https://github.com/ballerina-platform/ballerina-lang/issues/8678


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/OCSPStaplingHandler.java
```diff
@@ -50,33 +52,40 @@ protected boolean verify(ChannelHandlerContext ctx, ReferenceCountedOpenSslEngin
         //Get the stapled ocsp response from the ssl engine.
         byte[] staple = engine.getOcspResponse();
         if (staple == null) {
-            throw new IllegalStateException("Server didn't provide an OCSP staple!");
-        }
+            // If the response came from the server does not contain the OCSP staple, client attempts to validate
+            // the certificate by directly calling OCSP access location and if that also fails, finally
+            // do the CRL validation.
+            int cacheSize = 50;
+            int cacheDelay = 15;
```
Manuri: Are these default values? Should these be configurable?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/OCSPStaplingHandler.java
```diff
@@ -50,33 +52,40 @@ protected boolean verify(ChannelHandlerContext ctx, ReferenceCountedOpenSslEngin
...
+            RevocationVerificationManager revocationVerifier = new RevocationVerificationManager(cacheSize, cacheDelay);
+            return revocationVerifier.verifyRevocationStatus(engine.getSession().getPeerCertificateChain());
+        } else {
```
chamil321: Shall we not specifically bring an else scope as the code has returned from the IF scope?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/OCSPStaplingHandler.java
Bhashinee: These are default values for OCSP and CRL. Since the client is configured to connect through OCSP stapling, user does not need to configure them. So we are using default values. This will be executed only when the server fails to send the OCSP staple.

PR #171, opened 2018-05-22
Move channel reset and remove message future after the LastHttpContent

chamil321:

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/MessageFuture.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/MessageFuture.java
@@ -45,12 +45,14 @@ public class MessageFuture {
                 notifyMessageListener(httpContent);
                 if (httpContent instanceof LastHttpContent) {
                     this.httpCarbonMessage.removeMessageFuture();
+                    return;
                 }
             }
             while (!pendingPayload.isEmpty()) {
                 notifyMessageListener(pendingPayload.poll());
                 if (pendingPayload.poll() instanceof LastHttpContent) {
                     this.httpCarbonMessage.removeMessageFuture();
+                    return;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/MessageFuture.java
```diff
@@ -36,24 +37,26 @@ public MessageFuture(HTTPCarbonMessage httpCarbonMessage) {
...
-    public synchronized void setMessageListener(MessageListener messageListener) {
-        this.messageListener = messageListener;
+    public void setMessageListener(MessageListener messageListener) {
         synchronized (httpCarbonMessage) {
+            this.messageListener = messageListener;
             while (!httpCarbonMessage.isEmpty()) {
                 HttpContent httpContent = httpCarbonMessage.getHttpContent();
                 notifyMessageListener(httpContent);
+                if (httpContent instanceof LastHttpContent) {
+                    this.httpCarbonMessage.removeMessageFuture();
```
shafreenAnfar: Shall we just "return" from here. Because for a given message there is only one LastHttpContent.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/MessageFuture.java
```diff
@@ -36,24 +37,26 @@ public MessageFuture(HTTPCarbonMessage httpCarbonMessage) {
...
+                }
             }
             while (!pendingPayload.isEmpty()) {
                 notifyMessageListener(pendingPayload.poll());
+                if (pendingPayload.poll() instanceof LastHttpContent) {
+                    this.httpCarbonMessage.removeMessageFuture();
```
shafreenAnfar: Shall we just "return" from here. Because for a given message there is only one LastHttpContent.

chamil321: Fixed with ae03d5b

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/MessageFuture.java
chamil321: Fixed with ae03d5b

PR #182, opened 2018-05-31
Reduce the number of ports used in test cases

Bhashinee:
## Purpose
Reduce the number of ports used in test cases


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/util/TestUtil.java
```diff
@@ -80,6 +80,9 @@
 
     public static final int HTTP_SERVER_PORT = 9000;
     public static final int HTTPS_SERVER_PORT = 9004;
+    public static final int SERVER_PORT1 = 9091;
+    public static final int SERVER_PORT2 = 9092;
+    public static final int SERVER_PORT3 = 9093;
```
shafreenAnfar: Shall we change these to 9001, 9002, 9003

PR #188, opened 2018-06-04
[WebSocket] Fix on error notification issue

irunika:


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/websocket/client/WebSocketTestClientConnectorListener.java
```diff
@@ -216,6 +217,13 @@ public WebSocketCloseMessage getCloseMessage() {
         return closeMessage;
     }
 
+    /**
+     * Return the throwable available
```
chamil321: Shall we use either Return available throwable or Return available error ?

PR #197, opened 2018-06-13
Implement trailer headers support in HTTP/2

### LINKED ISSUE #181: Add trailer header support
**Description:**
Current http2 implementation doesn't support passing trailer header frame at the end of the stream. We need to support this to support gRPC over http2. because in gRPC, it is mandated to have trailer header in server response. It carries request status and end of stream flag.

**Suggested Labels:**
improvement

**Related Issues:**
ballerina-platform/ballerina-lang#8460

daneshk:
## Purpose
Resolves #181

## Goals
Implement trailer headers support in HTTP/2.

## Approach
If we need to send trailer headers, we need to add them to LastHttpContent and add content to carbon message. Handler checks for trailer headers in LastHttpContent. If exists, it sends trailers as separate headers with end stream flag.

## User stories
Enable sending trailer headers in carbon message.

## Release note
Implement trailer headers support in HTTP/2.


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/http2/Http2SourceHandler.java
```diff
@@ -121,19 +124,24 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
...
-            HTTPCarbonMessage sourceReqCMsg = setupHttp2CarbonMsg(headersFrame.getHeaders(), streamId);
             if (headersFrame.isEndOfStream()) {
-                // Add empty last http content if no data frames available in the http request
-                sourceReqCMsg.addHttpContent(new DefaultLastHttpContent());
+                // Retrieve HTTP request and add last http content with trailer headers.
+                HTTPCarbonMessage sourceReqCMsg = streamIdRequestMap.get(streamId);
+                if (sourceReqCMsg != null) {
+                    readTrailerHeaders(streamId, headersFrame.getHeaders(), sourceReqCMsg);
+                    streamIdRequestMap.remove(streamId);
+                }
```
isudana: What will happen if this header frame is the initial header frame and also it has endOfStream ?

daneshk: So shall we check whether HTTP method exists in the header and create the message. 
Trailer header doesn't contain any HTTP method. So we cannot create the message from the trailer header.

isudana: +1

daneshk: Fixed eee9803c34f9904d00da260ec733c9b2698aeddb

PR #206, opened 2018-06-18
Allow terminating a connection with a close frame

riyafa:

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/DefaultWebSocketConnection.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/DefaultWebSocketConnection.java
@@ -167,10 +167,35 @@ public class DefaultWebSocketConnection implements WebSocketConnection {
 
     @Override
     public ChannelFuture terminateConnection() {
+        frameHandler.setCloseInitialized(true);
         return ctx.close();
     }
 
-    public int getCloseInitiatedStatusCode() {
+    @Override
+    public ChannelFuture terminateConnection(int statusCode, String reason) {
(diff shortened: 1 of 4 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/DefaultWebSocketConnection.java
```diff
@@ -167,9 +167,34 @@ public ChannelFuture finishConnectionClosure(int statusCode, String reason) {
...
     public ChannelFuture terminateConnection() {
+        frameHandler.setCloseInitialized(true);
         return ctx.close();
     }
 
+    @Override
+    public ChannelFuture terminateConnection(int statusCode, String reason) {
+        ChannelPromise closePromise = ctx.newPromise();
+        ctx.writeAndFlush(new CloseWebSocketFrame(statusCode, reason)).addListener(
+                writeFuture -> {
```
irunika: This can go inline with the upper line

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/DefaultWebSocketConnection.java
```diff
@@ -167,9 +167,34 @@ public ChannelFuture finishConnectionClosure(int statusCode, String reason) {
...
+                    frameHandler.setCloseInitialized(true);
+                    Throwable writeCause = writeFuture.cause();
+                    if (writeFuture.isSuccess() && writeCause != null) {
+                        closePromise.setFailure(writeCause);
+                        ctx.close();
+                    } else {
```
irunika: Isn't it much cleaner to use return rather than using else everywhere?

PR #222, opened 2018-07-03
Fix response for HEAD method

irunika:
## Automation tests
 - Unit tests 
   > Yes
 - Integration tests
   > No

## Security checks

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
@@ -24,6 +24,7 @@ import io.netty.buffer.Unpooled;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http.DefaultLastHttpContent;
 import io.netty.handler.codec.http.HttpContent;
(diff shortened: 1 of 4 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
```diff
@@ -144,15 +147,16 @@ private void writeOutboundResponse(HTTPCarbonMessage outboundResponseMsg, boolea
...
+
                 } else {
                     contentLength += httpContent.content().readableBytes();
                     setupContentLengthRequest(outboundResponseMsg, contentLength);
                     outboundChannelFuture = writeOutboundResponseHeaderAndBody(outboundResponseMsg,
                             (LastHttpContent) httpContent, keepAlive);
                 }
             } else {
-                outboundChannelFuture = writeOutboundResponseBody(httpContent);
+                outboundChannelFuture = checkAndWriteLastHttpContent(httpContent);
```
shafreenAnfar: Can we come up with a better method name ? It writes more than LastHttpContent I guess.

irunika: It checks whether the original request is a header request and writes the `LastHttpContent` of the message. So should we say like `checkHeadRequestAndWriteLastHttpContent`?
Or we can say just `writeLastHttpContent`?
What would be the better approach?

PR #224, opened 2018-07-04
Add 100-continue specific states

chamil321:
> Address application layer problem of responding to a request which has not been read completely. In that instance, states are changed from RECEIVING_ENTITY_BODY to SENDING_ENTITY_BODY. Since user doesn't consume the inbound request payload, the connection is closed after checking this state inconsistent transition.


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Constants.java
```diff
@@ -119,6 +119,7 @@
     public static final String X_FORWARDED_BY = "x-forwarded-by";
     public static final String X_FORWARDED_HOST = "x-forwarded-host";
     public static final String X_FORWARDED_PROTO = "x-forwarded-proto";
+    public static final String CONTINUE = "100-continue";
```
irunika: Can we change the name of this? IMHO `CONTINUE` seems more generic. `CONTINUE_100` might be a better option.

PR #227, opened 2018-07-08
Refactor Sourcehandler/TargetHandler code

shafreenAnfar:
Please note that this does not have any functional changes.

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
@@ -44,17 +44,24 @@ import org.slf4j.LoggerFactory;
 import org.wso2.transport.http.netty.common.ssl.SSLConfig;
 import org.wso2.transport.http.netty.common.ssl.SSLHandlerFactory;
 import org.wso2.transport.http.netty.config.ChunkConfig;
+import org.wso2.transport.http.netty.config.KeepAliveConfig;
 import org.wso2.transport.http.netty.config.Parameter;
 import org.wso2.transport.http.netty.config.SslConfiguration;
 import org.wso2.transport.http.netty.contract.HttpResponseFuture;
+import org.wso2.transport.http.netty.listener.SourceHandler;
 import org.wso2.transport.http.netty.message.DefaultListener;
 import org.wso2.transport.http.netty.message.HTTPCarbonMessage;
+import org.wso2.transport.http.netty.message.HttpCarbonRequest;
+import org.wso2.transport.http.netty.message.HttpCarbonResponse;
 import org.wso2.transport.http.netty.message.Listener;
+import org.wso2.transport.http.netty.message.PooledDataStreamerFactory;
 import org.wso2.transport.http.netty.sender.CertificateValidationHandler;
 import org.wso2.transport.http.netty.sender.OCSPStaplingHandler;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.nio.channels.ClosedChannelException;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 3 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
```diff
@@ -740,4 +748,109 @@ public static void safelyRemoveHandlers(ChannelPipeline pipeline, String... hand
...
+            } else {
+                return false;
+            }
+        case ALWAYS:
+            return true;
+        case NEVER:
+            return false;
+        default:
+            // The execution will never reach here. To make the code compilable, default case has to be placed here.
+            return true;
```
Manuri: If execution should never reach here, better throw an exception rather than returning true. Then if someone adds another value to the enum(mistakenly may be) it will be caught here.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/SourceHandler.java
```diff
@@ -100,100 +95,60 @@ public SourceHandler(ServerConnectorFuture serverConnectorFuture, String interfa
...
                     }
+                } catch (RuntimeException ex) {
+                    httpContent.release();
+                    log.warn("Response already received before completing the inbound request" + ex.getMessage());
                 }
-            } else {
-                log.warn("Inconsistent state detected : inboundRequestMsg is null for channel read event");
             }
+        } else {
+            log.warn("Inconsistent state detected : inboundRequestMsg is null for channel read event");
```
irunika: Shouldn't we notify error and close the channel here since we have received a corrupted HTTP request?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/TargetHandler.java
```diff
@@ -139,25 +94,51 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
+            inboundResponseMsg.addHttpContent(httpContent);
+
+            if (isLastHttpContent(httpContent)) {
+                this.inboundResponseMsg = null;
+                targetChannel.getChannel().pipeline().remove(Constants.IDLE_STATE_HANDLER);
+                targetErrorHandler.setState(ENTITY_BODY_RECEIVED);
+                if (!isKeepAlive(keepAliveConfig, outboundRequestMsg)) {
+                    closeChannel(ctx);
+                }
+
```
irunika: Can we remove the new line here?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/Util.java
```diff
@@ -740,4 +748,109 @@ public static void safelyRemoveHandlers(ChannelPipeline pipeline, String... hand
...
+        if (ctx.channel().pipeline().get(Constants.SSL_HANDLER) != null) {
+            isSecuredConnection = true;
+        }
+        inboundRequestMsg.setProperty(Constants.IS_SECURED_CONNECTION, isSecuredConnection);
+
+        inboundRequestMsg.setProperty(Constants.LOCAL_ADDRESS, ctx.channel().localAddress());
+        inboundRequestMsg.setProperty(Constants.REMOTE_ADDRESS, remoteAddress);
+        inboundRequestMsg.setProperty(Constants.REQUEST_URL, httpRequestHeaders.uri());
+        inboundRequestMsg.setProperty(Constants.TO, httpRequestHeaders.uri());
+        //Added protocol name as a string
```
riyafa: Shouldn't this comment be above line 793?

VijithaEkanayake: We can remove that comment

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/SourceHandler.java
VijithaEkanayake: I think warn log is more appropriate and we don't need to close the connection as there is no spec violation kind of thing here.

PR #230, opened 2018-07-19
Fix SonarQube issues

riyafa:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/SourceHandler.java
```diff
@@ -141,15 +141,15 @@ private void readInboundReqEntityBody(Object inboundRequestEntityBody) throws Se
                     }
                 } catch (RuntimeException ex) {
                     httpContent.release();
-                    log.warn("Response already received before completing the inbound request" + ex.getMessage());
+                    log.warn("Response already received before completing the inbound request.", ex);
```
chamil321: Shall we log only the message of exception? ex.getMessage()*

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/sender/TargetErrorHandler.java
```diff
@@ -115,19 +115,19 @@ protected void handleErrorIdleScenarios(HTTPCarbonMessage inboundResponseMsg, St
...
+    private void handleIncompleteInboundResponse(HttpCarbonMessage inboundResponseMsg, String errorMessage) {
         LastHttpContent lastHttpContent = new DefaultLastHttpContent();
         lastHttpContent.setDecoderResult(DecoderResult.failure(new DecoderException(errorMessage)));
         inboundResponseMsg.addHttpContent(lastHttpContent);
         log.warn(errorMessage);
     }
 
     public void exceptionCaught(Throwable cause) {
-        log.warn("Exception occurred in TargetHandler : " + cause.getMessage());
+        log.warn("Exception occurred in TargetHandler : ", cause);
```
chamil321: Shall we log only the message here as well? cause.getMessage()*

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/certificatevalidation/crl/CRLVerifier.java
```diff
@@ -136,10 +134,6 @@ protected X509CRL downloadCRLFromWeb(String crlURL) throws IOException, Certific
             throw new CertificateVerificationException(e);
         } catch (CRLException e) {
             throw new CertificateVerificationException("Cannot generate X509CRL from the stream data", e);
-        } finally {
-            if (crlStream != null) {
-                crlStream.close();
```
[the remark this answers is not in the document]
riyafa: Using try with resources. It is insisted by sonar Qube

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/certificatevalidation/ocsp/OCSPVerifier.java
```diff
@@ -165,12 +165,11 @@ public static OCSPResp getOCSPResponce(String serviceUrl, OCSPReq request) throw
                 connection.setRequestProperty("Accept", "application/ocsp-response");
                 connection.setDoOutput(true);
                 try (OutputStream out = connection.getOutputStream();
-                        DataOutputStream dataOut = new DataOutputStream(new BufferedOutputStream(out));) {
+                        DataOutputStream dataOut = new DataOutputStream(new BufferedOutputStream(out))) {
 
                     dataOut.write(array);
 
                     dataOut.flush();
-                    dataOut.close();
```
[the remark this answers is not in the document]
riyafa: Because:  "close()" calls should not be redundant. Try with resources handles the close.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/certificatevalidation/cache/CacheManager.java
```diff
@@ -34,7 +34,7 @@
  */
 public class CacheManager {
 
-    private final boolean doNotInterruptIfRunning = false;
+    private static final boolean DO_NOT_INTERRUPT_IF_RUNNING = false;
```
irunika: Why did you make this static? Which means each method changes a class level variable. Not instance level. Is it the expected behaviour?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/DefaultWebSocketConnection.java
```diff
@@ -34,6 +35,7 @@
     private int closeInitiatedStatusCode;
     private String id;
     private String negotiatedSubProtocol;
+    private static final String MESSAGE_QUEUE_HANDLER = "MESSAGE_QUEUE_HANDLER";
```
irunika: Why have we moved this from Constants to here? AFAIR this is used in multiple classes.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/WebSocketInboundFrameHandler.java
```diff
@@ -178,13 +172,10 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
         } else if (msg instanceof ContinuationWebSocketFrame) {
             ContinuationWebSocketFrame frame = (ContinuationWebSocketFrame) msg;
-            switch (continuationFrameType) {
-                case TEXT:
-                    notifyTextMessage(frame, frame.text(), frame.isFinalFragment());
-                    break;
-                case BINARY:
-                    notifyBinaryMessage(frame, frame.content(), frame.isFinalFragment());
-                    break;
+            if (continuationFrameType == WebSocketFrameType.TEXT) {
```
irunika: Special reason to change this to `if` statement?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/DefaultWebSocketConnection.java
irunika: IMO this constant should remain in `Constants`.  Because by design message queue handler is not a part of WebSocket. It can queue any message type. So `MESSAGE_QUEUE_HANDLER` should be a constant that everyone can see. So it will not make confuses when reusing MessageQueueHandler.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/certificatevalidation/cache/CacheManager.java
riyafa: SonarQube: Public constants and fields initialized at declaration should be "static final" rather than merely "final"
Making a public constant just final as opposed to static final leads to duplicating its value for every instance of the class, uselessly increasing the amount of memory required to execute the application.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/DefaultWebSocketConnection.java
riyafa: yes. `MESSAGE_QUEUE_HANDLER` is usable by anyone, but they can give it any name they like when adding to pipeline. In WebSocket usecase this constant is used in only this class. I have ensured pipeline is modified by only methods in only this class and not elsewhere to reduce complexities in code.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/certificatevalidation/cache/CacheManager.java
irunika: I get it. My question here is that was this used by instance level or class level. By the code, it seems we can remove this constant since this is used only for a boolean and we can directly use this there since this implementation is not going to change.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/WebSocketInboundFrameHandler.java
riyafa: It is for code readability. For only 2 checks speed is not **the** concern

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/common/certificatevalidation/cache/CacheManager.java
riyafa: It is at the command of the owner. I am fixing sonarQubes in this PR. Maybe it gives more clarity to use a constant even for a boolean.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/websocket/DefaultWebSocketConnection.java
irunika: Yes I get your point. What I was thinking here is that this constant does not have anything relates to WebSocket. So if someone is using the message queue again in the code this constant can be used if we make it public. So a single copy of this variable can be used in many places rather than specify it in WebSocketConnection. Here, only WebSocketConnection can use the single copy of this variable rather than using it in the whole project.

riyafa: Moved it to `Constants`

PR #247, opened 2018-09-02
Handle multipart request with expect:100-continue header

chamil321:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/listener/states/listener/Expect100ContinueHeaderReceived.java
```diff
@@ -69,7 +74,10 @@ public void readInboundRequestHeaders(HttpCarbonMessage inboundRequestMsg, HttpR
 
     @Override
     public void readInboundRequestBody(Object inboundRequestEntityBody) throws ServerConnectorException {
-        log.warn("readInboundRequestBody {}", ILLEGAL_STATE_ERROR);
+        // Some abnormal clients send request body without waiting for the 100-continue response.
```
arukshani: Calling this an "abnormal" client is not right I think coz according to the spec,

1) A client that sends a 100-continue expectation is not required to wait for any specific length of time; such a client MAY proceed to send the message body even if it has not yet received a response.

2)  A server MAY omit sending a 100 (Continue) response if it has already received some or all of the message body for the corresponding request, or if the framing indicates that there is no message body.

https://tools.ietf.org/html/rfc7231#section-5.1.1

So shall we remove the abnormal part from the comment as it gives a wrong impression?

PR #261, opened 2018-09-27
Fix idle timeout to use only ALL_IDLE

riyafa:
## Purpose
Makes sure connection is closed only for ALL_IDLE event. Earlier the connection was closed for READER_IDLE or WRITER_IDLE events as well.

Resolve https://github.com/ballerina-platform/ballerina-lang/issues/10381

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/listener/HttpServerChannelInitializer.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/listener/HttpServerChannelInitializer.java
@@ -213,8 +213,7 @@ public class HttpServerChannelInitializer extends ChannelInitializer<SocketChann
                                        this.pipeliningNeeded, this.pipeliningLimit));
         if (socketIdleTimeout >= 0) {
             serverPipeline.addBefore(Constants.HTTP_SOURCE_HANDLER, Constants.IDLE_STATE_HANDLER,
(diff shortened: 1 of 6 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/listener/HttpServerChannelInitializer.java
```diff
@@ -213,8 +213,7 @@ public void configureHttpPipeline(ChannelPipeline serverPipeline, String initial
                                        this.pipeliningNeeded, this.pipeliningLimit));
         if (socketIdleTimeout >= 0) {
             serverPipeline.addBefore(Constants.HTTP_SOURCE_HANDLER, Constants.IDLE_STATE_HANDLER,
-                    new IdleStateHandler(socketIdleTimeout, socketIdleTimeout, socketIdleTimeout,
-                            TimeUnit.MILLISECONDS));
+                                     new IdleStateHandler(0, 0, socketIdleTimeout, TimeUnit.MILLISECONDS));
```
wggihan: Don't we have any use case to alter this value based on a situation? Is this correct always pass the 0?

riyafa: The value `0` is used to disable the handler for the triggering user events when reader or writer are idle. We have only one configuration for timeout in the Ballerina side. This is supposed to close the connection when both reader and writer are idle. Therefore we set the value for only allIdle because this is the only user event we are interested in.
We don't have any use cases for reader or writer idle values currently.

PR #262, opened 2018-09-28
Respond properly for unsupported WebSocket version

riyafa:

### DIFF
--- a/components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/websocket/server/WebSocketServerHandshakeFunctionalityTestCase.java
+++ b/components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/websocket/server/WebSocketServerHandshakeFunctionalityTestCase.java
@@ -18,6 +18,11 @@
 
 package org.wso2.transport.http.netty.websocket.server;
 
+import com.mashape.unirest.http.HttpResponse;
+import com.mashape.unirest.http.Unirest;
+import com.mashape.unirest.http.exceptions.UnirestException;
+import io.netty.handler.codec.http.HttpHeaderNames;
+import io.netty.handler.codec.http.HttpHeaderValues;
 import io.netty.handler.codec.http.websocketx.CloseWebSocketFrame;
 import io.netty.handler.codec.http.websocketx.WebSocketHandshakeException;
 import org.testng.Assert;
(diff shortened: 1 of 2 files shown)


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/websocket/server/WebSocketServerHandshakeFunctionalityTestCase.java
```diff
@@ -86,6 +91,18 @@ public void testSuccessfulHandshake() throws URISyntaxException, InterruptedExce
         testClient.closeChannel();
     }
 
+    @Test(description = "Check response for unsupported WebSocket version")
+    public void testUnsupportedWebSocketVersion() throws UnirestException {
+        String url = String.format("http://%s:%d/%s", TestUtil.TEST_HOST, TestUtil.SERVER_CONNECTOR_PORT, "test");
+        HttpResponse<String> response = Unirest.get(url).header(HttpHeaderNames.UPGRADE.toString(),
```
wggihan: Shall we check the possibility of improving the readability of this statement? May be assigned to variables and using those?

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/websocket/server/WebSocketServerHandshakeFunctionalityTestCase.java
```diff
@@ -86,6 +91,19 @@ public void testSuccessfulHandshake() throws URISyntaxException, InterruptedExce
         testClient.closeChannel();
     }
 
+    @Test(description = "Check response for unsupported WebSocket version")
+    public void testUnsupportedWebSocketVersion() throws UnirestException {
+        String url = String.format("http://%s:%d/%s", TestUtil.TEST_HOST, TestUtil.SERVER_CONNECTOR_PORT, "test");
+        HttpResponse<String> response = Unirest.get(url)
+                .header(HttpHeaderNames.UPGRADE.toString(), HttpHeaderValues.WEBSOCKET.toString())
```
chamil321: Shall we static import these constants?

riyafa: These constants are only used in one location in the whole class. Static imports are encouraged to be used only sparingly and that is when a constant is used in multiple locations in the code: https://docs.oracle.com/javase/1.5.0/docs/guide/language/static-import.html

chamil321: Documentation explains mainly two point IIUC. 
1. Use static imports only when you were tempted to "abuse inheritance".
2. Import individual members.

My comment is about the second point. Importing individual members rather than whole class improves the readability as it removes the redundant class qualification. 
Same is mentioned in the documentation as well, "Used appropriately, static import can make your program more readable, by removing the boilerplate of repetition of class names."

riyafa: I have not changed it because it says "use it when you require frequent access to static members from one or two classes." Here it is only a single access.

PR #266, opened 2018-10-08
Add outbound throttling

riyafa:
## Purpose

Client can go OOM if the server is slow during a send operation because of data collecting in the netty buffer. Similarly server can go OOM if the client is slow during a respond operation. Outbound throttling fixes this issue.

The following scenarios require outbound throttling and are addressed in this PR:
- Server responding with a large file.
- Client sending a large file to server.
- Passthrough of a large request file.
- Passthrough of a large response file.

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
@@ -26,6 +26,8 @@ import org.slf4j.LoggerFactory;
 import org.wso2.transport.http.netty.contract.HttpConnectorListener;
 import org.wso2.transport.http.netty.contract.config.ChunkConfig;
 import org.wso2.transport.http.netty.contract.config.KeepAliveConfig;
+import org.wso2.transport.http.netty.contractimpl.common.BackPressureHandler;
+import org.wso2.transport.http.netty.contractimpl.common.Util;
 import org.wso2.transport.http.netty.contractimpl.common.states.MessageStateContext;
 import org.wso2.transport.http.netty.contractimpl.listener.RequestDataHolder;
 import org.wso2.transport.http.netty.contractimpl.listener.SourceHandler;
@@ -69,23 +71,25 @@ public class HttpOutboundRespListener implements HttpConnectorListener {
 
     @Override
     public void onMessage(HttpCarbonMessage outboundResponseMsg) {
-        sourceContext.channel().eventLoop().execute(() -> {
+        BackPressureHandler backpressureHandler = Util.getBackPressureHandler(sourceContext);
+        Util.setBackPressureListener(outboundResponseMsg.isPassthrough(), backpressureHandler,
+                                     outboundResponseMsg.getTargetContext());
+        if (handlerExecutor != null) {
+            handlerExecutor.executeAtSourceResponseReceiving(outboundResponseMsg);
+        }
 
-            if (handlerExecutor != null) {
-                handlerExecutor.executeAtSourceResponseReceiving(outboundResponseMsg);
-            }
-
-            outboundResponseMsg.getHttpContentAsync().setMessageListener(httpContent ->
-                    this.sourceContext.channel().eventLoop().execute(() -> {
-                        try {
-                            writeOutboundResponse(outboundResponseMsg, httpContent);
-                        } catch (Exception exception) {
-                            String errorMsg = "Failed to send the outbound response : "
-                                    + exception.getMessage().toLowerCase(Locale.ENGLISH);
-                            LOG.error(errorMsg, exception);
-                            inboundRequestMsg.getHttpOutboundRespStatusFuture().notifyHttpListener(exception);
-                        }
-                    }));
+        outboundResponseMsg.getHttpContentAsync().setMessageListener(httpContent -> {
+            Util.checkUnWritabilityAndNotify(sourceContext, backpressureHandler);
+            this.sourceContext.channel().eventLoop().execute(() -> {
+                try {
+                    writeOutboundResponse(outboundResponseMsg, httpContent);
+                } catch (Exception exception) {
+                    String errorMsg = "Failed to send the outbound response : "
+                            + exception.getMessage().toLowerCase(Locale.ENGLISH);
+                    LOG.error(errorMsg, exception);
+                    inboundRequestMsg.getHttpOutboundRespStatusFuture().notifyHttpListener(exception);
+                }
+            });
         });
     }
 
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
@@ -16,6 +16,7 @@
 package org.wso2.transport.http.netty.contractimpl.common;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
@@ -52,11 +53,13 @@ import org.wso2.transport.http.netty.contractimpl.common.ssl.SSLHandlerFactory;
 import org.wso2.transport.http.netty.contractimpl.listener.SourceHandler;
 import org.wso2.transport.http.netty.contractimpl.sender.CertificateValidationHandler;
 import org.wso2.transport.http.netty.contractimpl.sender.OCSPStaplingHandler;
+import org.wso2.transport.http.netty.message.DefaultBackPressureListener;
 import org.wso2.transport.http.netty.message.DefaultListener;
 import org.wso2.transport.http.netty.message.HttpCarbonMessage;
 import org.wso2.transport.http.netty.message.HttpCarbonRequest;
 import org.wso2.transport.http.netty.message.HttpCarbonResponse;
 import org.wso2.transport.http.netty.message.Listener;
+import org.wso2.transport.http.netty.message.PassthroughBackPressureListener;
 import org.wso2.transport.http.netty.message.PooledDataStreamerFactory;
 
 import java.io.IOException;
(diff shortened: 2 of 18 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
```diff
@@ -69,23 +72,30 @@ public HttpOutboundRespListener(HttpCarbonMessage requestMsg, SourceHandler sour
...
+            outboundThrottlingHandler.getOutboundThrottlingObservable().setListener(
+                    new DefaultOutboundThrottlingListener(sourceContext));
+        }
+        if (handlerExecutor != null) {
+            handlerExecutor.executeAtSourceResponseReceiving(outboundResponseMsg);
+        }
+
+        outboundResponseMsg.getHttpContentAsync().setMessageListener(httpContent -> {
+            if (outboundThrottlingHandler != null) {
+                //This is in the Ballerina thread
```
arukshani: Shall we remove this comment? This is not always the case.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
```diff
@@ -69,23 +72,30 @@ public HttpOutboundRespListener(HttpCarbonMessage requestMsg, SourceHandler sour
...
     @Override
     public void onMessage(HttpCarbonMessage outboundResponseMsg) {
-        sourceContext.channel().eventLoop().execute(() -> {
-
-            if (handlerExecutor != null) {
-                handlerExecutor.executeAtSourceResponseReceiving(outboundResponseMsg);
+        OutboundThrottlingHandler outboundThrottlingHandler = Util.getOutBoundThrottlingHandler(sourceContext);
+        if (outboundThrottlingHandler != null) {
+            outboundThrottlingHandler.getOutboundThrottlingObservable().setListener(
+                    new DefaultOutboundThrottlingListener(sourceContext));
```
arukshani: We need to handle the passthrough case as well. Can you please add that part

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/channel/TargetChannel.java
```diff
@@ -194,17 +210,22 @@ public void writeContent(HttpCarbonMessage httpOutboundRequest) {
...
-                    } catch (Exception exception) {
-                        String errorMsg = "Failed to send the request : "
-                                + exception.getMessage().toLowerCase(Locale.ENGLISH);
-                        LOG.error(errorMsg, exception);
-                        this.targetHandler.getHttpResponseFuture().notifyHttpListener(exception);
-                    }
-                })));
+        httpOutboundRequest.getHttpContentAsync().setMessageListener((httpContent -> {
+            if (outboundThrottlingHandler != null) {
+                //This is in the Ballerina thread
```
arukshani: Again this is not always the case. So shall we remove this comment?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/PassthroughOutboundThrottlingListener.java
```diff
@@ -0,0 +1,64 @@
...
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default implementation of the {@link OutboundThrottlingListener}.
+ */
+public class PassthroughOutboundThrottlingListener implements OutboundThrottlingListener {
+    private static final Logger LOG = LoggerFactory.getLogger(HttpMessageDataStreamer.class);
```
Bhashinee: Is it the wrong class passed to get the Logger?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/OutboundThrottlingHandler.java
```diff
@@ -0,0 +1,54 @@
...
+import io.netty.channel.ChannelInboundHandlerAdapter;
+import org.wso2.transport.http.netty.message.DefaultOutboundThrottlingObservable;
+import org.wso2.transport.http.netty.message.OutboundThrottlingObservable;
+
+/**
+ * Handles outbound throttling.
+ * Overrides the channelWritabilityChanged method to check the writability of the channel which is needed for
+ * outbound throttling.
+ */
+public class OutboundThrottlingHandler extends ChannelInboundHandlerAdapter {
```
arukshani: Would `BackPressureHandler` be a more suitable name for this handler? `OutboundThrottlingHandler` is quite a mouthful.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
```diff
@@ -809,4 +809,16 @@ public static void setHostNameVerfication(SSLEngine sslEngine) {
...
         sslEngine.setSSLParameters(sslParams);
     }
+
+    /**
+     * Use this method to get the {@link BackPressureHandler} in the pipeline. This requires a
+     * {@link ChannelHandlerContext} of a handler in the pipeline.
+     *
+     * @param channelContext the channelContext which will be used to obtain the {@link BackPressureHandler}
+     *                       in the pipeline.
+     */
```
chamil321: Document the return value too.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/channel/TargetChannel.java
```diff
@@ -180,6 +184,18 @@ public Http2ClientChannel getHttp2ClientChannel() {
     }
 
     public void writeContent(HttpCarbonMessage httpOutboundRequest) {
+        BackPressureHandler backpressureHandler = Util.getBackPressureHandler(
+                targetHandler.getContext());
```
chamil321: Shall we move this code piece to above line?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/channel/TargetChannel.java
```diff
@@ -180,6 +184,18 @@ public Http2ClientChannel getHttp2ClientChannel() {
...
+        if (backpressureHandler != null) {
+            if (httpOutboundRequest.getSourceContext() != null && httpOutboundRequest.isPassthrough()) {
```
chamil321: Shall we move this if-else scope to another modular function to reduce the complexity of writeContent method? Same code might not get reused, but it improves the readability.
We can call is as setBackPressureObserver() or setBackPressureObservableListener()

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/HttpCarbonMessage.java
```diff
@@ -446,7 +450,41 @@ public void setPipeliningFuture(HttpPipeliningFuture pipeliningFuture) {
...
+
+    /**
+     * @return true if it is a passthrough(when message body is not built).
+     */
+    public boolean isPassthrough() {
+        return passthrough;
+    }
+
+    /**
+     * This value is to be set when the message is to be sent to the consumer without building/processing the message
```
chamil321: Shall we rephrase this like "This value is set when sending the message without building/processing to the consumer in the application layer."

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/MessageFuture.java
```diff
@@ -47,6 +47,12 @@ public void setMessageListener(MessageListener messageListener) {
                     return;
                 }
             }
+
+            //Removes Inbound throttling listener during passthrough so that only backpressure handling would be
```
chamil321: Shall we add a space after //

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/PassthroughBackPressureListener.java
```diff
@@ -0,0 +1,64 @@
...
+    private Channel outChannel;
+
+    /**
+     * Sets the incoming and outgoing message channels.
+     *
+     * @param inContext This will be used to block and resume read interest of the incoming channel.
+     * @param outContext This will be used to check the writability of the outgoing channel.
+     */
+    public PassthroughBackPressureListener(ChannelHandlerContext inContext,
+                                           ChannelHandlerContext outContext) {
```
chamil321: Shall we move the content to above line?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/PassthroughBackPressureListener.java
```diff
@@ -0,0 +1,64 @@
...
+
+package org.wso2.transport.http.netty.message;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Default implementation of the {@link BackPressureListener}.
```
chamil321: Shall we mention pass-through relevance in the description?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/HttpOutboundRespListener.java
```diff
@@ -69,23 +73,35 @@ public HttpOutboundRespListener(HttpCarbonMessage requestMsg, SourceHandler sour
 
     @Override
     public void onMessage(HttpCarbonMessage outboundResponseMsg) {
-        sourceContext.channel().eventLoop().execute(() -> {
-
-            if (handlerExecutor != null) {
-                handlerExecutor.executeAtSourceResponseReceiving(outboundResponseMsg);
+        BackPressureHandler backpressureHandler = Util.getBackPressureHandler(sourceContext);
+        if (backpressureHandler != null) {
+            if (outboundResponseMsg.getTargetContext() != null && outboundResponseMsg.isPassthrough()) {
```
chamil321: Shall we move this if-else scope to another modular function to reduce the complexity of onMessage method? Same code might not get reused, but it improves the readability.
We can call is as setBackPressureObserver() or setBackPressureObservableListener()

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/HttpCarbonMessage.java
```diff
@@ -446,7 +450,41 @@ public void setPipeliningFuture(HttpPipeliningFuture pipeliningFuture) {
...
     }
 
+    /**
+     * Removes the content listener that is set for handling Inbound throttling.
+     */
     public void removeInboundContentListener() {
         this.contentObservable.removeListener();
     }
+
+    /**
```
VijithaEkanayake: I think is better to add descriptions in the java doc comments. Even these methods are self-descriptive, Better to add doc comment with a broader context. Just documenting only the return values and the params seem bit odd.

riyafa: Fixed

arukshani: the thing is this is not only used for throttling but for data binding scenario as well. So specifying the usage is not necessary IMO. It just removes the content listener and it is upto the  code that calling this has to give a context as to why that removal is necessary

VijithaEkanayake: Agree, What I meant to say is the methods that have already documented with java doc comments which don't have a description rather it documented only the params or the return values. It is fine to avoid java docs comments in the places where it requires callers context but leaving java doc comments without description is odd.

arukshani: yes but this method neither takes a param nor returns a value. Therefore shall we remove the java doc as this is self-explanatory.

arukshani: my mistake. I have confused this with the above method. Please ignore.

PR #270, opened 2018-10-18
Release not consuming outbound request contents

chamil321:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/states/EntityBodyReceived.java
```diff
@@ -43,7 +43,9 @@ public void writeOutboundRequestHeaders(HttpCarbonMessage httpOutboundRequest, H
 
     @Override
     public void writeOutboundRequestEntity(HttpCarbonMessage httpOutboundRequest, HttpContent httpContent) {
-        LOG.warn("writeOutboundRequestEntity {}", ILLEGAL_STATE_ERROR);
+        // This method might get called when inbound response is received but outbound response is not yet completed.
```
VijithaEkanayake: shouldn't this be an outbound request?

PR #273, opened 2018-10-26
Introduce a different thread pool to manage pipelined response serialization

arukshani:
Fixes https://github.com/ballerina-platform/ballerina-lang/issues/10878


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/listener/HttpServerChannelInitializer.java
```diff
@@ -212,7 +214,7 @@ public void configureHttpPipeline(ChannelPipeline serverPipeline, String initial
         serverPipeline.addLast(Constants.HTTP_SOURCE_HANDLER,
                                new SourceHandler(this.serverConnectorFuture, this.interfaceId, this.chunkConfig,
                                                  keepAliveConfig, this.serverName, this.allChannels,
-                                       this.pipeliningNeeded, this.pipeliningLimit));
+                                       this.pipeliningNeeded, this.pipeliningLimit, this.pipeliningGroup));
```
Manuri: Whats the different between "Needed" and "Enabled"? Just wondering because we have "oscpStaplingEnabled" but "pipeliningNeeded".

arukshani: We can go ahead with enabled. I have fixed it in all places. Thanks for pointing out :)

PR #277, opened 2018-12-07
Add proper error message when the host cannot be resolved

arukshani:
$subject


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/ClientConnectorUnknownHostTestCase.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
VijithaEkanayake: Shouldn't this year 2018??

arukshani: :)

PR #279, opened 2018-12-12
Add config for maxFrameSize

riyafa:

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/websocket/WebSocketClientConnectorConfig.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/websocket/WebSocketClientConnectorConfig.java
@@ -39,6 +39,7 @@ public class WebSocketClientConnectorConfig extends SslConfiguration {
     private int idleTimeoutInSeconds;
     private boolean autoRead;
     private final HttpHeaders headers;
+    private int maxFrameSize = 65536;
 
     public WebSocketClientConnectorConfig(String remoteAddress) {
         this.remoteAddress = remoteAddress;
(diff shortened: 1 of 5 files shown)


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/util/server/websocket/WebSocketRemoteServerInitializer.java
```diff
@@ -43,13 +43,13 @@ public WebSocketRemoteServerInitializer(SslContext sslCtx, String subProtocols)
...
     @Override
-    public void initChannel(SocketChannel ch) throws Exception {
+    public void initChannel(SocketChannel ch) {
         ChannelPipeline pipeline = ch.pipeline();
         if (sslCtx != null) {
             pipeline.addLast(sslCtx.newHandler(ch.alloc()));
         }
         pipeline.addLast(new HttpServerCodec());
-        pipeline.addLast(new HttpObjectAggregator(65536));
+        pipeline.addLast(new HttpObjectAggregator(8192));
```
arukshani: Shall we define `8192` as a constant and use it?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/websocket/WebSocketClientConnectorConfig.java
```diff
@@ -39,6 +39,7 @@
     private int idleTimeoutInSeconds;
     private boolean autoRead;
     private final HttpHeaders headers;
+    private int maxFrameSize = 65536;
```
Manuri: Shouldn't this be a final constant?

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/util/server/websocket/WebSocketRemoteServerInitializer.java
Manuri: Can we make this value also a constant indicating what this means?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/websocket/WebSocketClientConnectorConfig.java
riyafa: It is the default value for this parameter. This parameter is not a constant.

PR #282, opened 2018-12-14
Support setting mutual ssl as 'optional'

Bhashinee:
## Purpose
Fix the issue: https://github.com/ballerina-platform/ballerina-lang/issues/12801


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
```diff
@@ -57,6 +58,8 @@ public void setKeyStorePass(String keyStorePassword) {
     public void setVerifyClient(String verifyClient) {
         if (REQUIRE.equalsIgnoreCase(verifyClient)) {
             sslConfig.setNeedClientAuth(true);
+        } else if (OPTIONAL.equalsIgnoreCase(verifyClient)) {
+            sslConfig.setWantClientAuth(true);
         }
```
shafreenAnfar: Better to have else and log something.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/ssl/SSLHandlerFactory.java
```diff
@@ -130,7 +132,11 @@ private KeyStore getKeyStore(File keyStore, String keyStorePassword) throws IOEx
     public SSLEngine buildServerSSLEngine(SSLContext sslContext) {
         SSLEngine engine = sslContext.createSSLEngine();
         engine.setUseClientMode(false);
-        engine.setNeedClientAuth(needClientAuth);
+        if (needClientAuth) {
+            engine.setNeedClientAuth(true);
+        } else if (wantClientAuth) {
+            engine.setWantClientAuth(true);
+        }
```
shafreenAnfar: Here also better to have a else and log a message.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
```diff
@@ -716,6 +717,8 @@ public static HttpCarbonMessage createInboundReqCarbonMsg(HttpRequest httpReques
         inboundRequestMsg.setProperty(Constants.REMOTE_ADDRESS, sourceHandler.getRemoteAddress());
         inboundRequestMsg.setProperty(Constants.REQUEST_URL, httpRequestHeaders.uri());
         inboundRequestMsg.setProperty(Constants.TO, httpRequestHeaders.uri());
+        inboundRequestMsg.setProperty(MUTUAL_SSL_HANDSHAKE_RESULT,
+                ctx.channel().attr(Constants.MUTUAL_SSL_RESULT_ATTRIBUTE).get());
```
chamil321: A potential Null pointer ? ctx.channel().attr(Constants.MUTUAL_SSL_RESULT_ATTRIBUTE)
Can we guarantee that always (sslEngine.getWantClientAuth() || sslEngine.getNeedClientAuth()) condition is true?

Bhashinee: If the condition is false it will not throw a null pointer exception. It will set the property value as null.

chamil321: Doing .get() call on null value will cause the exception.

chamil321: Doing get() call on null value will cause the exception.

chamil321: Since the null check is done at ballerina level and there is no any default value to be set instead, this is fine.

PR #290, opened 2019-01-23
Verify certificate expiry date

Bhashinee:
## Purpose
Fix the issue: https://github.com/ballerina-platform/ballerina-lang/issues/13255

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
@@ -330,10 +330,9 @@ public class Util {
 
             if (referenceCountedOpenSslContext != null) {
                 SslHandler sslHandler = referenceCountedOpenSslContext.newHandler(socketChannel.alloc());
-                ReferenceCountedOpenSslEngine engine = (ReferenceCountedOpenSslEngine) sslHandler.engine();
-                sslEngine = engine;
+                sslEngine = sslHandler.engine();
                 socketChannel.pipeline().addLast(sslHandler);
-                socketChannel.pipeline().addLast(new OCSPStaplingHandler(engine));
+                socketChannel.pipeline().addLast(new OCSPStaplingHandler((ReferenceCountedOpenSslEngine) sslEngine));
             }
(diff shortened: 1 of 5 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
```diff
@@ -330,11 +331,11 @@ public static void configureHttpPipelineForSSL(SocketChannel socketChannel, Stri
             if (referenceCountedOpenSslContext != null) {
                 SslHandler sslHandler = referenceCountedOpenSslContext.newHandler(socketChannel.alloc());
                 ReferenceCountedOpenSslEngine engine = (ReferenceCountedOpenSslEngine) sslHandler.engine();
+                sslEngine = engine;
```
Manuri: Can't we directly assign to sslEngine variable in the previous line and then parse it line 336?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/listener/SslHandshakeCompletionHandlerForServer.java
```diff
@@ -57,7 +61,9 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
             if (event.isSuccess()) {
                 if (sslEngine.getWantClientAuth() || sslEngine.getNeedClientAuth()) {
                     try {
-                        sslEngine.getSession().getPeerCertificates();
+                        Certificate[] certs = sslEngine.getSession().getPeerCertificates();
+                        X509Certificate endUserCert = (X509Certificate) certs[0];
```
Manuri: Possible ArrayIndexOutOfBoundsException?

Bhashinee: It will not throw an ArrayIndexOutOfBoundsException. Because if an SSL session has a successful handshake with the peer, there should be at least one peer certificate. Otherwise, it will throw SSLPeerUnverifiedException.

PR #291, opened 2019-01-24
Facilitate the use of shared connection manager among HTTP clients

arukshani:
## Automation tests
 - Unit tests 
   > Added

## Security checks


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/channel/pool/ConnectionManager.java
```diff
@@ -206,4 +97,58 @@ public void invalidateTargetChannel(TargetChannel targetChannel) throws Exceptio
...
+        GenericObjectPool clientPool;
+        synchronized (this) {
+            if (!this.connGlobalPool.containsKey(httpRoute.toString())) {
+                PoolableTargetChannelFactory poolableTargetChannelFactory = new PoolableTargetChannelFactory(group,
+                    eventLoopClass, httpRoute, senderConfig, bootstrapConfig, this);
+                clientPool = createPoolForRoute(poolableTargetChannelFactory);
+                this.connGlobalPool.put(httpRoute.toString(), clientPool);
+            } else {
+                clientPool = this.connGlobalPool.get(httpRoute.toString());
+            }
```
Manuri: We can optimize this by 
1. Checking for key existence outside the synchronized block and return the client pool if it exists already.
2. Then inside the synchronized block, check for key existence again, create and return a pool if key doesn't exists or just return the pool if the key exists.
Otherwise even when a pool exists, threads will always block here.

arukshani: I already tried that. But under heavy load it throws error, if we do that change. So I left it as it is. Double-checked locking doesn't seem to work always. Anyway, this will be accessed only once because once we get the pool, we keep a reference to it in client connector, so for each and every request this wont get called.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/channel/pool/ConnectionManager.java
```diff
@@ -206,4 +97,58 @@ public void invalidateTargetChannel(TargetChannel targetChannel) throws Exceptio
...
+     * Gets the client connection pool.
+     *
+     * @param httpRoute        Represents the endpoint address
+     * @param sourceHandler    Represents the source channel
+     * @param senderConfig     Represents the client configurations
+     * @param bootstrapConfig  Represents the bootstrap info related to client connection creation
+     * @param clientEventGroup Represents the eventloop group that the client channel should be bound to
+     * @return the client connection pool
+     */
+    public GenericObjectPool getClientPool(HttpRoute httpRoute, SourceHandler sourceHandler,
```
daneshk: Shall we rename the method to getClientConnectionPool? It looks more self descriptive. wdyt?

PR #298, opened 2019-02-13
Provide getFullHttpCarbonMessage API to read message with complete content

chamil321:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/DefaultFullHttpMessageFuture.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+
+    @Override
+    public synchronized void removeListener() {
+        messageListener = null;
+    }
+
+    @Override
+    public void notifySuccess() {
```
shafreenAnfar: Shall we add a small note saying why this is not synchronized ?

chamil321: Added the description

PR #301, opened 2019-02-20
Fix HTTP/2 error rate that gets occurred under high load

arukshani:
## Security checks

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
@@ -160,6 +160,8 @@ public class DefaultHttpClientConnector implements HttpClientConnector {
             }
         }
 
+        //Cannot directly assign srcHandler and http2SourceHandler to inner class ConnectionAvailabilityListener hence
+        //need two new separate variables
         final SourceHandler h1SourceHandler = srcHandler;
         final Http2SourceHandler h2SourceHandler = http2SourceHandler;
 
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
@@ -42,6 +42,7 @@ import org.wso2.transport.http.netty.contract.config.SenderConfiguration;
 import org.wso2.transport.http.netty.contractimpl.common.HttpRoute;
 import org.wso2.transport.http.netty.contractimpl.common.ssl.SSLConfig;
 import org.wso2.transport.http.netty.contractimpl.listener.SourceHandler;
+import org.wso2.transport.http.netty.contractimpl.listener.http2.Http2SourceHandler;
 import org.wso2.transport.http.netty.contractimpl.sender.ConnectionAvailabilityListener;
 import org.wso2.transport.http.netty.contractimpl.sender.channel.BootstrapConfiguration;
 import org.wso2.transport.http.netty.contractimpl.sender.channel.TargetChannel;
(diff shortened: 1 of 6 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
```diff
@@ -147,8 +148,24 @@ public HttpResponseFuture send(HttpCarbonMessage httpOutboundRequest) {
...
+            if (sourceHandlerObject instanceof SourceHandler) {
+                srcHandler = (SourceHandler) sourceHandlerObject;
+            } else if (sourceHandlerObject instanceof Http2SourceHandler) {
+                http2SourceHandler = (Http2SourceHandler) sourceHandlerObject;
+            }
+        }
+
+        //Cannot directly assign srcHandler and http2SourceHandler to inner class ConnectionAvailabilityListener hence
+        //need two new separate variables
+        final SourceHandler h1SourceHandler = srcHandler;
```
VijithaEkanayake: Shall we use http1xSrcHandlder and http2SrcHandler for the variables names?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
```diff
@@ -189,10 +207,23 @@ public void onSuccess(String protocol, ChannelFuture channelFuture) {
                         LOG.debug("Created the connection to address: {}",
                                   route.toString() + " " + "Original Channel ID is : " + channelFuture.channel().id());
                     }
-                    if (srcHandler != null) {
+
+                    if (protocol.equalsIgnoreCase(Constants.HTTP_SCHEME) && h1SourceHandler != null) {
```
VijithaEkanayake: Shall we switch constant and the variable here to avoid possible NPEs??
```suggestion
                    if ((Constants.HTTP_SCHEME).equalsIgnoreCase(protocol) && h1SourceHandler != null) {
```

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
```diff
@@ -176,7 +193,8 @@ public HttpResponseFuture send(OutboundMsgHolder outboundMsgHolder, HttpCarbonMe
             }
 
             // Look for the connection from http connection manager
-            TargetChannel targetChannel = connectionManager.borrowTargetChannel(route, srcHandler, senderConfiguration,
+            TargetChannel targetChannel = connectionManager.borrowTargetChannel(route, srcHandler, http2SourceHandler,
```
shafreenAnfar: AFAIU, code snippet from 151 to 168, can be moved to here. It makes the code easier to read as closely associated code get grouped together.

shafreenAnfar: Also, I think we should come up with two borrowTargetChannel methods. Probably something as follows, 

public TargetChannel borrowHttpTargetChannel()
public TargetChannel borrowHttp2TargetChannel()

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
```diff
@@ -189,10 +207,23 @@ public void onSuccess(String protocol, ChannelFuture channelFuture) {
                         LOG.debug("Created the connection to address: {}",
                                   route.toString() + " " + "Original Channel ID is : " + channelFuture.channel().id());
                     }
-                    if (srcHandler != null) {
+
+                    if (Constants.HTTP_SCHEME.equalsIgnoreCase(protocol) && http1xSrcHandlder != null) {
```
shafreenAnfar: We can remove these if/else conditions, If we use three different listeners. AFAIU, that would make the code simpler.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
shafreenAnfar: I will do these changes :)

shafreenAnfar: I will do these changes :)

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
shafreenAnfar: I will do these changes :)

PR #303, opened 2019-02-26
Support setting ssl handshake/session time out

Bhashinee:
## Purpose
Currently, there is no way of changing the default values of SSL session time out and SSL handshake time out. This PR is to introduce the capability to set SSL handshake and session time outs as a user prefers.

Fixed issues: https://github.com/ballerina-platform/ballerina-lang/issues/13847
https://github.com/ballerina-platform/ballerina-lang/issues/13846

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
@@ -68,6 +68,7 @@ import java.nio.channels.ClosedChannelException;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import javax.net.ssl.SSLEngine;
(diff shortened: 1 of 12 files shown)


components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/http2/Http2ALPNwithCertsTest.java
```diff
@@ -73,6 +73,8 @@ private ListenerConfiguration getListenerConfigs() {
         listenerConfiguration.setPort(TestUtil.SERVER_PORT1);
         listenerConfiguration.setScheme(HTTPS_SCHEME);
         listenerConfiguration.setVersion(String.valueOf(HTTP_2_0));
+        listenerConfiguration.setSslSessionTimeOut(30);
+        listenerConfiguration.setSslHandshakeTimeOut(20);
```
riyafa: Shall we use a test constant for these values as they are used in all classes?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
```diff
@@ -872,4 +877,10 @@ public static void checkUnWritabilityAndNotify(ChannelHandlerContext context,
             }
         }
     }
+
+    public static void setSslHandshakeTimeOut(SSLConfig sslConfig, SslHandler sslHandler) {
+        if (sslConfig.getHandshakeTimeOut() > 0) {
+            sslHandler.setHandshakeTimeout(sslConfig.getHandshakeTimeOut(), TimeUnit.SECONDS);
+        }
+    }
```
riyafa:
```suggestion
int handshakeTimeout = sslConfig.getHandshakeTimeOut();
if ( handshakeTimeout > 0) {
            sslHandler.setHandshakeTimeout(handshakeTimeout, TimeUnit.SECONDS);
        }
    }
```

PR #310, opened 2019-03-15
Remove HTTP/2 client connection from the pool when the stream ids are exhausted 

arukshani:
$subject. 

Fixes https://github.com/ballerina-platform/ballerina-lang/issues/14380


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/http2/Http2TargetHandler.java
```diff
@@ -146,6 +147,11 @@ void writeContent(ChannelHandlerContext ctx) {
                     http2ClientChannel.getChannel().eventLoop().execute(() -> {
                         try {
                             writeOutboundRequest(ctx, httpContent);
+                        } catch (Http2NoMoreStreamIdsException ex) {
+                            //Remove connection from the pool
+                            http2ClientChannel.removeFromConnectionPool();
+                            LOG.warn("Channel is removed from the connection pool : ", ex.getMessage(), ex);
```
ldclakmal:
```suggestion
                            LOG.warn("Channel is removed from the connection pool : {}", ex.getMessage(), ex);
```

PR #317, opened 2019-04-01
Fix Netty SSL exception on server startup

lasanthaS:
## Purpose
Fix https://github.com/wso2/transport-http/issues/314

## Security checks


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/listener/SslHandshakeCompletionHandlerForServer.java
```diff
@@ -76,6 +80,11 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
                 ctx.close();
             }
         }
+     }
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+        LOG.warn("Error while SSL handshake: " + cause.getMessage());
```
Bhashinee: Shall we make this warn to an error? 
```suggestion
        LOG.error("Error while SSL handshake: " + cause.getMessage());
```
I think it would be better to log it as an error.

PR #325, opened 2019-04-11
Prevent waiting when shutting down the connector factory

minudika:
## Security checks


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpWsConnectorFactory.java
```diff
@@ -169,4 +169,19 @@ public void shutdown() throws InterruptedException {
             pipeliningGroup.shutdownGracefully().sync();
         }
     }
+
+    /**
+     * This method is for shutting down the connectors without a delay.
+     *
+     * @throws InterruptedException when interrupted by some other event
+     */
+    public void shutdownNow() throws InterruptedException {
```
chamil321: Let's remove this exception throw as the sync() is not applied

minudika: Fixed with 74907e9c2fedc26db36e423b96c835d58cccf71c

PR #326, opened 2019-04-25
Refactor the code to separate http1x code from http2 

shafreenAnfar:

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
@@ -77,7 +77,7 @@ public class DefaultHttpClientConnector implements HttpClientConnector {
     private String httpVersion;
     private ChunkConfig chunkConfig;
     private KeepAliveConfig keepAliveConfig;
-    private boolean isHttp2;
+    private boolean http2;
     private ForwardedExtensionConfig forwardedExtensionConfig;
     private EventLoopGroup clientEventGroup;
     private BootstrapConfiguration bootstrapConfig;
@@ -89,10 +89,19 @@ public class DefaultHttpClientConnector implements HttpClientConnector {
         this.senderConfiguration = senderConfiguration;
         initTargetChannelProperties(senderConfiguration);
         if (Float.valueOf(senderConfiguration.getHttpVersion()) == Constants.HTTP_2_0) {
-            isHttp2 = true;
+            http2 = true;
         }
         this.clientEventGroup = clientEventGroup;
         this.bootstrapConfig = bootstrapConfig;
+    }
+
+    private void initTargetChannelProperties(SenderConfiguration senderConfiguration) {
+        this.httpVersion = senderConfiguration.getHttpVersion();
+        this.chunkConfig = senderConfiguration.getChunkingConfig();
+        this.socketIdleTimeout = senderConfiguration.getSocketIdleTimeout(Constants.ENDPOINT_TIMEOUT);
+        this.sslConfig = senderConfiguration.getClientSSLConfig();
+        this.keepAliveConfig = senderConfiguration.getKeepAliveConfig();
+        this.forwardedExtensionConfig = senderConfiguration.getForwardedExtensionConfig();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


arukshani:
Shall we run a load test for H2-H1 and H1-H2 passthrough scenarios? We might need to check for both the h1 source handler and h2 source handler for a connection. If not we might again experience a low TPS for the mentioned scenarios.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
```diff
@@ -142,153 +151,108 @@ public boolean close() {
...
+        } catch (Exception failedCause) {
+            return notifyListenerAndGetErrorResponseFuture(failedCause);
         }
+    }
+
+    private HttpResponseFuture sendHttp2Request(OutboundMsgHolder outboundRequestHolder, HttpRoute route)
+            throws Exception {
+        HttpResponseFuture httpResponseFuture;
+
+        final Http2SourceHandler http2SourceHandler =
```
arukshani: We don't know whether the H2 client resides in an h1 or an h2 service or just inside a main hence we need to look for the connection in both the h1 and h2 source handler pool. borrowChannel() method needs both and that logic needs to change inside borrowChannel() too.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
```diff
@@ -142,153 +151,108 @@ public boolean close() {
...
-                    connectionManager.getHttp2ConnectionManager().
-                        addHttp2ClientChannel(freshHttp2ClientChannel.getChannel().eventLoop(), route,
-                                              freshHttp2ClientChannel);
-                    freshHttp2ClientChannel.addDataEventListener(Constants.IDLE_STATE_HANDLER,
-                                                                 new TimeoutHandler(socketIdleTimeout,
-                                                                                    freshHttp2ClientChannel));
-                    new RequestWriteStarter(outboundMsgHolder, freshHttp2ClientChannel).startWritingContent();
-                    httpResponseFuture.notifyResponseHandle(new ResponseHandle(outboundMsgHolder));
-                }
+        SourceHandler srcHandler = getSourceHandler(outboundRequestHolder.getRequestCarbonMessage());
```
arukshani: Same goes for the H1 client as well. What if it resides inside H2 service. Need the same logic change as mentioned in the previous comment.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
```diff
@@ -379,4 +334,137 @@ private void setProxyAuthorizationHeader(HttpCarbonMessage httpOutboundRequest)
...
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Created the connection to address: {}",
+                          route.toString() + " " + "Original Channel ID is : " + channelFuture.channel().id());
+            }
+
+            if (Constants.HTTP_SCHEME.equalsIgnoreCase(protocol) && http1xSrcHandler != null) {
+                channelFuture.channel().deregister()
+                        .addListener(future -> http1xSrcHandler.getEventLoop().register(channelFuture.channel())
+                                .addListener(future1 -> startExecutingOutboundRequest(protocol, channelFuture)));
+            } else {
```
arukshani: We need to check whether the successfully created connection is H1 and whether it is in HTTP/2 source handler as well. Otherwise, in H2-H1 passthrough scenario, we cannot guarantee that the incoming http content transfer will happen in the same thread as in the source.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/DefaultHttpClientConnector.java
```diff
@@ -379,4 +334,137 @@ private void setProxyAuthorizationHeader(HttpCarbonMessage httpOutboundRequest)
...
+                LOG.debug("Created the connection to address: {}",
+                          route.toString() + " " + "Original Channel ID is : " + channelFuture.channel().id());
+            }
+
+            if (Constants.HTTP_SCHEME.equalsIgnoreCase(protocol) && http2SrcHandler != null) {
+                channelFuture.channel().deregister()
+                        .addListener(future -> http2SrcHandler.getChannelHandlerContext().channel().eventLoop()
+                                .register(channelFuture.channel())
+                                .addListener(future1 -> startExecutingOutboundRequest(protocol, channelFuture)));
+            } else {
```
arukshani: Again if the connection is downgraded to H1 and if that H2 client is in H1 service, basically if in passthrough scenario this will be equivalent to H1-H1 and incoming content will not transfer to the backend via the same thread as in the source.

shafreenAnfar:
Hi All,

I am closing this PR based on Rukshan's feedback. Will send a new PR addressing those concerns.

PR #340, opened 2019-07-19
Provide an option to disable SSL certificate validation

Bhashinee:
## Purpose
Fix the issue: https://github.com/ballerina-platform/ballerina-lang/issues/15147


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
```diff
@@ -169,6 +169,10 @@ public void setSslHandshakeTimeOut(long handshakeTimeOut) {
         sslConfig.setHandshakeTimeOut(handshakeTimeOut);
     }
 
+    public void setDisableSsl(boolean disableSsl) {
```
chamil321: Do we require parameter? I think the boolean is always true.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
```diff
@@ -236,26 +240,29 @@ private SSLConfig getSSLConfigForListener() {
...
 
     private SSLConfig getSSLConfigForSender() {
-        if ((sslConfig.getTrustStore() == null || sslConfig.getTrustStorePass() == null) && (
-                sslConfig.getClientTrustCertificates() == null)) {
-            throw new IllegalArgumentException("TrustStoreFile or TrustStorePassword not defined for HTTPS/WS scheme");
-        }
-        if (sslConfig.getTrustStore() != null) {
-            if (!sslConfig.getTrustStore().exists()) {
-                throw new IllegalArgumentException("TrustStore File " + sslConfig.getTrustStore() + " not found");
+        if (!sslConfig.isDisableSsl()) {
```
chamil321: Shall we check the inverse of this condition and return?
```suggestion
        if (sslConfig.isDisableSsl()) {
               return sslConfig;
        }
```

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/disablessl/SslDisabledClientTest.java
```diff
@@ -0,0 +1,95 @@
...
+        senderConfiguration.setDisableSsl(true);
+        senderConfiguration.setScheme(HTTPS_SCHEME);
+        return senderConfiguration;
+    }
+
+    @Test
+    public void testSslDisabledClient() {
+        TestUtil.testHttpsPost(httpClientConnector, TestUtil.SERVER_PORT3);
+    }
+    @AfterClass
```
chamil321: Leave a new line

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
wggihan: Since we are not setting any value, shall we use some any other name other than setDisableSsl? WDYT?

Bhashinee: Changed it to 'disableSsl()'

PR #341, opened 2019-07-26
Handle HTTP/2 client timeout based on the state the stream is in

arukshani:
$subject

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/states/http2/SenderState.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/states/http2/SenderState.java
@@ -93,9 +93,8 @@ public interface SenderState {
     /**
      * Handles stream timeout.
      *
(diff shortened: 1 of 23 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/sender/states/http2/SenderState.java
```diff
@@ -89,4 +89,13 @@ void readInboundResponseBody(ChannelHandlerContext ctx, Http2DataFrame http2Data
...
                             OutboundMsgHolder outboundMsgHolder);
+
+    /**
+     * Handles stream timeout.
+     *
+     * @param ctx               the channel handler context
+     * @param outboundMsgHolder the outbound message holder
+     * @param serverPush        indicates the response type
+     */
+    void handleStreamTimeout(ChannelHandlerContext ctx, OutboundMsgHolder outboundMsgHolder, boolean serverPush);
```
chamil321: Does ChannelHandlerContext ctx required as a parameter?

PR #344, opened 2019-08-06
Remove the redundant thread contention

shafreenAnfar:


arukshani:
Just to be on the safe side shall we do a small test by sending a large file with respond() call(not a passthrough) with the other side having a slow client to process that response, to see that when the back pressure kicks in it doesn't block the I/O thread?

shafreenAnfar:
I think this is not applicable anymore. Hence closing it.

PR #363, opened 2019-08-20
Send 100-continue response with HTTP/2 server when expect header is present

arukshani:
$subject


arukshani:
This feature needs tests. Only manually tested. 

To write test we need to extend netty http/2 client since java 8 client doesn't support HTTP/2 and our HTTP/2 client doesn't handle 100-continue responses

PR #364, opened 2019-08-21
Remove the chance to happen deadlock in the webSocket

kalaiyarasiganeshalingam:
## Purpose
if we use sync() , 
- waits for this future until it is done, and rethrows the cause of the failure if this future failed.
- executes an infinite loop, doesn't return from a call to request some resource. 

So I use addlistener() insteadof sync(). 

## Automation tests
 - Unit tests 
   > Code coverage information
 - Integration tests
   > Details about the test cases and coverage

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no


riyafa:
Shall we provide a description on how this prevents any deadlock?

PR #373, opened 2019-09-03
Remove the default status code value

shafreenAnfar:


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/listener/states/Expect100ContinueHeaderReceived.java
```diff
@@ -88,7 +89,7 @@ public void writeOutboundResponseHeaders(HttpCarbonMessage outboundResponseMsg,
     @Override
     public void writeOutboundResponseBody(HttpOutboundRespListener outboundResponseListener,
                                           HttpCarbonMessage outboundResponseMsg, HttpContent httpContent) {
-        if (outboundResponseMsg.getHttpStatusCode() == HttpResponseStatus.CONTINUE.code()) {
+        if (Util.getHttpResponseStatus(outboundResponseMsg).code() == HttpResponseStatus.CONTINUE.code()) {
```
chamil321: Shall we have a static method in Util class to check the condition as it is used redundantly?
```
    public static boolean continueResponse(HttpCarbonMessage outboundResponseMsg) {
        return getHttpResponseStatus(outboundResponseMsg).code() == HttpResponseStatus.CONTINUE.code();
    }
```

shafreenAnfar: True. But wondering if having a static is the right thing.

PR #378, opened 2019-09-05
Fix thread hanging issue

shafreenAnfar:

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/DefaultBackPressureListener.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/DefaultBackPressureListener.java
@@ -35,28 +35,28 @@ public class DefaultBackPressureListener implements BackPressureListener {
      * Creates the semaphore and sets the source or target channel.
      */
     public DefaultBackPressureListener() {
-
         this.semaphore = new Semaphore(0);
     }
 
     @Override
     public void onUnWritable() {
-
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("Semaphore acquired in thread {} ", Thread.currentThread().getName());
-        }
         try {
             semaphore.acquire();
         } catch (InterruptedException e) {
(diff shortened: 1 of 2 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/HttpCarbonMessage.java
```diff
@@ -450,6 +450,7 @@ public synchronized IOException getIoException() {
 
     public synchronized void setIoException(IOException ioException) {
         this.ioException = ioException;
+        contentObservable.removeListener();
```
arukshani: Shall we do this only for HTTP/1 like this.

String httpVersion = this.getHttpVersion();
        if (Constants.HTTP_1_1_VERSION.equalsIgnoreCase(httpVersion)) {
            contentObservable.removeListener();
        }

Can't remember exactly what, but with HTTP/2 flow controlling we had a problem with removeListener. So shall we do this only for HTTP/1 only for now. then if this works with http.2 also we can remove that check

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/message/DefaultBackPressureListener.java
```diff
@@ -35,28 +35,28 @@
...
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("Semaphore acquired in thread {} ", Thread.currentThread().getName());
-        }
         try {
             semaphore.acquire();
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
         }
+
+        if (LOG.isDebugEnabled()) {
```
arukshani: If we put the log here, wont this get printed once the semaphore is released? since the acquire gets blocked before before printing the log?

shafreenAnfar: Good catch. Yeah you are right. Fixed it.

PR #390, opened 2019-10-01
Write a test case for server closing the connection during the ssl handshake

Bhashinee:
## Purpose
To test the scenario of server closing the tcp connection during the SSL handshake.


shafreenAnfar:
Shall we reformat with given-when-then format.

PR #400, opened 2020-02-01
Add HTTP/1.1 Trailer support

chamil321:
## Automation tests
 - Unit tests 
   > Added
 - Integration tests
   > N/A


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/listener/states/SendingEntityBody.java
```diff
@@ -274,4 +274,12 @@ private void resetOutboundListenerState() {
         contentLength = 0;
         headersWritten = false;
     }
+
+    private void writeTrailerHeaders(LastHttpContent lastContent) {
+        if (lastContent == LastHttpContent.EMPTY_LAST_CONTENT) {
+            return;
+        }
```
daneshk: I guess we don't need to check whether the content is empty here. Even though the content is empty; the trailing headers can go with the object

chamil321: Ideally we should not get an EmptyLastContent. I found it when writing a test case for an echo listener which send a GET request by setting EmptyLastContent. I fixed the echo listener properly to send back a DefaultLastHttpContent. Hence removed the check via e2dd6595eb6f3f54078a34bdf8c9b82d0a62b59a

PR #405, opened 2020-03-10
Pass the mutual ssl handshake result to ballerina

Bhashinee:
## Purpose
$subject
Fix the issue: https://github.com/ballerina-platform/ballerina-lang/issues/21381

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/states/Http2StateUtil.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/states/Http2StateUtil.java
@@ -70,6 +70,7 @@ import static org.wso2.transport.http.netty.contract.Constants.INBOUND_REQUEST;
 import static org.wso2.transport.http.netty.contract.Constants.LISTENER_INTERFACE_ID;
 import static org.wso2.transport.http.netty.contract.Constants.LISTENER_PORT;
 import static org.wso2.transport.http.netty.contract.Constants.LOCAL_ADDRESS;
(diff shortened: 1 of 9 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/states/Http2StateUtil.java
```diff
@@ -144,6 +145,8 @@ public static HttpCarbonRequest setupCarbonRequest(HttpRequest httpRequest, Http
         sourceReqCMsg.setProperty(LISTENER_PORT, localAddress != null ? localAddress.getPort() : null);
         sourceReqCMsg.setProperty(LISTENER_INTERFACE_ID, http2SourceHandler.getInterfaceId());
         sourceReqCMsg.setProperty(PROTOCOL, HTTP_SCHEME);
+        sourceReqCMsg.setProperty(MUTUAL_SSL_HANDSHAKE_RESULT,
+                ctx.channel().attr(Constants.MUTUAL_SSL_RESULT_ATTRIBUTE).get());
```
chamil321: This expression can be null when `sslEngine.getWantClientAuth() || sslEngine.getNeedClientAuth()` condition is false. So at the ballerina level, a null check is needed. We can avoid it by setting it to some default value unless you perform a certain logic when `MUTUAL_SSL_HANDSHAKE_RESULT` is null.
WDYT?

Bhashinee: Added a default value as 'disabled'.

PR #411, opened 2020-06-18
Add support for sending HSTS headers for HTTPS

ramindu90:
Users have to configure HSTS header values in netty-transports.yaml under listener configurations as follows.

id: "msf4j-https"
....
scheme: https
....
strictTransportSecurityHeader:

The headers are added into the response only If the configuration values are not null and the listener
scheme is https. Configuration values in the netty-transport.yaml can be left blank if the headers are not
needed.

## Security checks

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/ListenerConfiguration.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/ListenerConfiguration.java
@@ -27,6 +27,9 @@ import java.util.List;
 public class ListenerConfiguration extends SslConfiguration {
 
     public static final String DEFAULT_KEY = "default";
+    public static final String DEFAULT_SCHEME = "https";
+    public static final String DEFAULT_HTTP_STRICT_TRANSPORT_SECURITY_HEADER_VALUE = "max-age=15768000;" +
+            " includeSubDomains";
 
     /**
      * @deprecated
@@ -35,7 +38,8 @@ public class ListenerConfiguration extends SslConfiguration {
     @Deprecated
     public static ListenerConfiguration getDefault() {
         ListenerConfiguration defaultConfig;
-        defaultConfig = new ListenerConfiguration(DEFAULT_KEY, "0.0.0.0", 8080);
+        defaultConfig = new ListenerConfiguration(DEFAULT_KEY, "0.0.0.0", 8080, DEFAULT_SCHEME,
+                DEFAULT_HTTP_STRICT_TRANSPORT_SECURITY_HEADER_VALUE);
         return defaultConfig;
     }
     private String id = DEFAULT_KEY;
@@ -55,6 +59,10 @@ public class ListenerConfiguration extends SslConfiguration {
     private boolean pipeliningEnabled;
     private boolean webSocketCompressionEnabled;
     private long pipeliningLimit;
+    private String strictTransportSecurityHeader;
+    private String keyStoreFile;
+    private String keyStorePass;
+    private String certPass;
 
     public ListenerConfiguration() {
     }
@@ -63,6 +71,14 @@ public class ListenerConfiguration extends SslConfiguration {
         this.id = id;
         this.host = host;
         this.port = port;
+    }
+
+    public ListenerConfiguration(String id, String host, int port, String scheme, String hstsHeader) {
+        this.id = id;
+        this.host = host;
+        this.port = port;
+        super.setScheme(scheme);
+        this.strictTransportSecurityHeader = hstsHeader;
     }
 
     public String getHost() {
@@ -200,5 +216,31 @@ public class ListenerConfiguration extends SslConfiguration {
 
     public void setWebSocketCompressionEnabled(boolean webSocketCompressionEnabled) {
         this.webSocketCompressionEnabled = webSocketCompressionEnabled;
+    }
+
+    public String getStrictTransportSecurityHeader() {
+        return strictTransportSecurityHeader;
+    }
+
+    public void setStrictTransportSecurityHeader(String strictTransportSecurityHeader) {
+        this.strictTransportSecurityHeader = strictTransportSecurityHeader;
+    }
+
+    public void setKeyStoreFile(String keyStoreFile) {
+        this.keyStoreFile = keyStoreFile;
+        super.setKeyStoreFile(this.keyStoreFile);
+    }
+
+    public void setKeyStorePass(String keyStorePass) {
+        this.keyStorePass = keyStorePass;
+        super.setKeyStorePass(this.keyStorePass);
+    }
+
+    public void setCertPass(String certPass) {
+        this.certPass = certPass;
+    }
+
+    public String getCertPass() {
+        return certPass;
     }
 }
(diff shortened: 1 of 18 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
```diff
@@ -197,7 +201,25 @@ private static void setOutboundRespHeaders(HttpCarbonMessage outboundResponseMsg
                                           ZonedDateTime.now().format(DateTimeFormatter.RFC_1123_DATE_TIME));
         }
 
-        outboundNettyResponse.headers().add(outboundResponseMsg.getHeaders());
+        Set<ListenerConfiguration> listenerConfigurations =
+                ConfigurationBuilder.getInstance().getConfiguration().getListenerConfigurations();
```
chamil321, line 205: When I check the `getConfiguration()` function, it performs a file read operation for netty-transports.yml file and load config to the `TransportsConfiguration`.

Also this method `setOutboundRespHeaders()` resides in the response path. therefore for every request this will get executed. So this may affect the performance

IMO ideally we should cache the file read and add a function to check only the config compatibility before setting the header.

chamil321: @shafreenAnfar Please add your thoughts.

shafreenAnfar: Yeah, this is completely wrong. We shouldn't get the listener configuration like this. Also, doing a I/O operation for each and every request/response transaction will drop performance drastically. As matter of fact, we shouldn't do any I/O operations in performance critical code :)

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/ListenerConfiguration.java
```diff
@@ -35,7 +38,8 @@
     @Deprecated
     public static ListenerConfiguration getDefault() {
         ListenerConfiguration defaultConfig;
-        defaultConfig = new ListenerConfiguration(DEFAULT_KEY, "0.0.0.0", 8080);
+        defaultConfig = new ListenerConfiguration(DEFAULT_KEY, "0.0.0.0", 8080, DEFAULT_SCHEME,
```
shafreenAnfar, line 41: Also this method is deprecated we need to remove this method.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SenderConfiguration.java
```diff
@@ -47,6 +47,8 @@ public static SenderConfiguration getDefault() {
     private String httpVersion = "1.1";
     private ProxyServerConfiguration proxyServerConfiguration;
     private PoolConfiguration poolConfiguration;
+    private String trustStoreFile;
```
shafreenAnfar, line 50: These should go to SslConfiugrtion because it is an attribute of that.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/ListenerConfiguration.java
```diff
@@ -55,6 +59,10 @@ public static ListenerConfiguration getDefault() {
     private boolean pipeliningEnabled;
     private boolean webSocketCompressionEnabled;
     private long pipeliningLimit;
+    private String strictTransportSecurityHeader;
```
shafreenAnfar, line 62: Why should the ListenereConfiugation should know about SSL stuff ?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
Bhashinee: As I got to know, this was a front-ported fix from https://github.com/wso2-support/transport-http/pull/5
We'll have to change that as well?

shafreenAnfar: Yeah

ramindu90: @shafreenAnfar moved the I/O operation at the initialization of the Util class in commit: d5512942b35f66ffe88c903578b32fe7dc96f131

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/ListenerConfiguration.java
ramindu90: @shafreenAnfar : Removed the method and done the necessary changes in commit: d5512942b35f66ffe88c903578b32fe7dc96f131

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SenderConfiguration.java
ramindu90: @shafreenAnfar : done in commit: d5512942b35f66ffe88c903578b32fe7dc96f131

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/ListenerConfiguration.java
ramindu90: @shafreenAnfar : moved the attributes to sslConfigurations in commit: d5512942b35f66ffe88c903578b32fe7dc96f131

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
```diff
@@ -131,6 +135,8 @@ private Util() {
     }
 
     private static final Logger LOG = LoggerFactory.getLogger(Util.class);
+    private static final Set<ListenerConfiguration>
+            listenerConfigurations = ConfigurationBuilder.getInstance().getConfiguration().getListenerConfigurations();
```
shafreenAnfar, line 139: Guys, this bit is not correct. Basically the transport code should only know about the listener configuration. How the listener configuration is built is part of whatever the application that uses it. Therefore, we shouldn't refer to any files from the transport code directly or indirectly.

anupama-pathirage:
@ramindu90 / @shafreenAnfar Is this still valid? Can we do the remaining suggested changes and merge it or close it if not valid? It seems this is too old to keep open.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/common/Util.java
ramindu90: @anupama-pathirage @shafreenAnfar Ack.

anupama-pathirage:
@ramindu90 Any update on this? If it takes  time to do the suggested changes shall we close this and re-open when it is ready?

PR #412, opened 2020-07-31
Fix the issue with HTTP2 content-encoding

Bhashinee:
## Purpose
Fix the https://github.com/ballerina-platform/ballerina-lang/issues/24977

## Goals
Fix the issue of HTTP2 server not encoding when the request is sent with multiple values in `accept-encoding` header.

## Approach
In HTTP1.1, this is done in [HttpContentCompressor](https://github.com/netty/netty/blob/0992718f87bf0f0c23660489b1f7e697a24f2a4b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java) by checking the `accept-encoding` header in this [determineWrapper](https://github.com/netty/netty/blob/0992718f87bf0f0c23660489b1f7e697a24f2a4b/codec-http/src/main/java/io/netty/handler/codec/http/HttpContentCompressor.java#L175) function. Same logic is applied in this PR, before setting the `content-encoding` header in the response. 

## Automation tests
 - Integration tests
   > An integration test was already there and fixed the test.

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
@@ -184,9 +184,9 @@ public class Http2OutboundRespListener implements HttpConnectorListener {
     }
 
     /***
-     * This function is to determine one encoding scheme from the request's `accept-encoding` header. This function is
-     * similar to Netty's `determineWrapper()` function in the `HttpContentCompressor` class which is used to do the
-     * same when doing the HTTP1.1 compression.
+     * This function is to determine one encoding scheme from the request's `accept-encoding` header. The logic to
+     * determine the scheme is similar to the logic used in Netty's `determineWrapper()` function in the
+     * `HttpContentCompressor` class which is used to do the same, when doing the HTTP1.1 compression.
      *
      * @param acceptEncoding `accept-encoding` header value
      * @return the chosen encoding scheme
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
@@ -51,6 +51,8 @@ import org.wso2.transport.http.netty.message.ServerRemoteFlowControlListener;
 import java.util.Calendar;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import static org.wso2.transport.http.netty.contract.Constants.ENCODING_DEFLATE;
+import static org.wso2.transport.http.netty.contract.Constants.ENCODING_GZIP;
 import static org.wso2.transport.http.netty.contract.Constants.PROMISED_STREAM_REJECTED_ERROR;
 import static org.wso2.transport.http.netty.contractimpl.common.states.Http2StateUtil.isValidStreamId;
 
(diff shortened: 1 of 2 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
```diff
@@ -171,9 +173,63 @@ private void setContentEncoding(HttpCarbonMessage outboundResponseMsg) {
...
+                    qValue = 0.0f;
+                }
+            }
+            if (encoding.contains("*")) {
+                starQ = qValue;
+            } else if (encoding.contains(ENCODING_GZIP) && qValue > gzipQ) {
+                gzipQ = qValue;
+            } else if (encoding.contains(ENCODING_DEFLATE) && qValue > deflateQ) {
+                deflateQ = qValue;
+            }
```
shafreenAnfar, line 215: Shall we have an else and log some error message. So that the code completes all the branches. IMO, there could be other custom compressions which we don't support.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/http2/compression/ResponseBodyCompressionTestCase.java
```diff
@@ -86,7 +86,7 @@ public void testResponseBodyCompression() throws Exception {
 
         streamId = streamId + 2;
         responseHandler = h2ClientWithoutDecompressor.sendPostRequest(PAYLOAD, streamId, "deflate;q=1.0, gzip;q=0.8");
-        assertCompressedResults("deflate;q=1.0, gzip;q=0.8", responseHandler.getFullResponse(streamId),
+        assertCompressedResults("deflate", responseHandler.getFullResponse(streamId),
```
shafreenAnfar, line 89: Can we add few more test scenarios here ? When adding need to add a test case to cover some custom compression scenario also.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
```diff
@@ -171,9 +173,63 @@ private void setContentEncoding(HttpCarbonMessage outboundResponseMsg) {
         if (contentEncoding == null) {
             String acceptEncoding = inboundRequestMsg.getHeader(HttpHeaderNames.ACCEPT_ENCODING.toString());
             if (acceptEncoding != null) {
-                outboundResponseMsg.setHeader(HttpHeaderNames.CONTENT_ENCODING.toString(), acceptEncoding);
+                // To set the `content-encoding` header, first we need to select a scheme based on the `q` value and
+                // server supporting encoding schemes when we have multiple values in the `accept-encoding` header.
```
shafreenAnfar, line 177: IMO, this comment is not needed. Code itself tells the story. If someone wants more info, he or she can read the spec.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/http2/compression/ResponseBodyCompressionTestCase.java
Bhashinee: According to the Spec, 

`If an Accept-Encoding header field is present in a request and none of the available representations for the response have a content-coding that is listed as acceptable, the origin server SHOULD send a response without any content-coding.`

So added a debug log instead an error log as the server is not obligated to use any compression method requested from the client.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
Bhashinee: According to the Spec, 

`If an Accept-Encoding header field is present in a request and none of the available representations for the response have a content-coding that is listed as acceptable, the origin server SHOULD send a response without any content-coding.`

So added a debug log instead an error log as the server is not obligated to use any compression method requested from the client.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
Bhashinee: Removed the comment.

components/org.wso2.transport.http.netty/src/test/java/org/wso2/transport/http/netty/http2/compression/ResponseBodyCompressionTestCase.java
Bhashinee: Added 3 more test cases for giving 
- not supporting encoding schemes
- `Identity` scheme
- Mixed collection of supporting and not supporting schemes
in the `Accept-Encoding` header.

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contractimpl/Http2OutboundRespListener.java
shafreenAnfar: Great!

PR #427, opened 2021-04-23
Add missing ssl configs and headers in listener

ramindu90:
Users have to configure HSTS header values in netty-transports.yaml under listener configurations as follows.

id: "msf4j-https"
....
scheme: https
....
strictTransportSecurityHeader: max-age=342352535; includeSubDomains
responseHeaderList: "Feature-Policy: microphone 'none'; geolocation 'none', Referrer-Policy: no-referrer"

Multiple headers can be added to responseHeaderList which can be separated by commas (",") . `responseHeaderList` and `strictTransportSecurityHeader` configuration values in the netty-transport.yaml can be left blank if the headers are not needed.

## Security checks

### DIFF
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
@@ -131,7 +131,7 @@ public class SslConfiguration {
         return String.valueOf(sslConfig.getKeyStore());
     }
 
-    public String getKeyStorePassword() {
+    public String getKeyStorePass() {
         return sslConfig.getKeyStorePass();
     }
 
--- a/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
+++ b/components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
@@ -49,6 +49,12 @@ public class SslConfiguration {
     private List<Parameter> parameters = new ArrayList<>();
     private SSLConfig sslConfig = new SSLConfig();
     private static final Logger LOG = LoggerFactory.getLogger(SslConfiguration.class);
+    private String keyStoreFile;
+    private String keyStorePassword;
+    private String certPass;
+    private String trustStoreFile;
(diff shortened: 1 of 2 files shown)


components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
```diff
@@ -285,4 +291,28 @@ private SSLConfig getSSLConfigForSender() {
...
+    public void setCertPass(String certPass) {
+        this.certPass = certPass;
+    }
+
+    public String getCertPass() {
+        return certPass;
+    }
+
+    public String getTrustStoreFile() {
+        return trustStoreFile;
```
chamil321, line 304: Don't we need setter for `trustStoreFile`?

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
```diff
@@ -285,4 +291,28 @@ private SSLConfig getSSLConfigForSender() {
...
+    }
+
+    public String getTrustStorePass() {
+        return trustStorePass;
```
chamil321, line 308: Same here

components/org.wso2.transport.http.netty/src/main/java/org/wso2/transport/http/netty/contract/config/SslConfiguration.java
```diff
@@ -49,6 +49,12 @@
     private List<Parameter> parameters = new ArrayList<>();
     private SSLConfig sslConfig = new SSLConfig();
     private static final Logger LOG = LoggerFactory.getLogger(SslConfiguration.class);
+    private String keyStoreFile;
+    private String keyStorePassword;
```
chamil321, line 53: These two class variable have not been used right?

ramindu90: These are added in due to yaml exceptions when reading Listener configuration in [deployment.yaml](https://github.com/wso2/analytics-apim/blob/master/product/distribution/carbon-home/conf/dashboard/deployment.yaml#L360)

chamil321 (approved):

